PR #16, opened 2022-01-14
Assignment 1

TechGlue:
No extra credit was attempted. The regular assignment is done.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -72,6 +72,22 @@ namespace PrincessBrideTrivia.Tests
             Assert.AreEqual(expectedString, percentage);
         }
 
+        [TestMethod]
+        public void RandomizeQuestions_GivenValidArray_ReturnsRandomizedArray()
+        {
+            //Arrange
+            Question[] questions = new Question[3];
+            Question[] questionsCpy = new Question[questions.Length];
+            
+            questions[0] = new Question {Text = "Question 0"};
+            questions[1] = new Question {Text = "Question 1"};
+            questions[2] = new Question {Text = "Question 2"};
+            questions.CopyTo(questionsCpy,0);
+            //Act 
+            questions = Program.RandomizeQuestions(questions);
+            //Assert
+            Assert.AreNotEqual(questionsCpy, questions);
+        }
 
         private static void GenerateQuestionsFile(string filePath, int numberOfQuestions)
         {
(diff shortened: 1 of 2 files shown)


TechGlue:
Extra Credit was attempted.

Ashrom0:
Will review this

Ashrom0 (commented):
Code in program and test seem to work well and seems that the extra credit also works

Pull Request title includes "Assignment 1" in the title ✔
Pull Request targets Assignment 1 branch ✔
Issue 1: Application no longer crashes ✔
Issue 2: Unit test properly passes ✔

Extra Credit
Issue 3: Feature request implemented ✔
Issue 3: Feature request unit tested ✔

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -72,6 +72,22 @@ public void GetFilePath_ReturnsFileThatExists()
...
+            Question[] questionsCpy = new Question[questions.Length];
+            
+            questions[0] = new Question {Text = "Question 0"};
+            questions[1] = new Question {Text = "Question 1"};
+            questions[2] = new Question {Text = "Question 2"};
+            questions.CopyTo(questionsCpy,0);
+            //Act 
+            questions = Program.RandomizeQuestions(questions);
+            //Assert
+            Assert.AreNotEqual(questionsCpy, questions);
```
mmwoodfo, line 89: This will always return false. Try putting this Assert before you randomize the questions, in doing this you'd expect the assert to fail, but it still passes.

A better option would be to use `CollectionAssert.AreNotEqual` and if you want to go the extra step, you can use `CollectionAssert.AreEquivalent` to make sure both arrays contain the same elements, regardless of their order.

Additionally, testing randomization is difficult. This test is not always guaranteed to pass. One way to get better odds of randomization is to do a few shuffles, so create one array that is the 'original' and then randomize it 5 times or so. Then check that each index is different between at least two of them, to prove that the questions do in fact move around. To seal the deal, you can do a `CollectionAssert.AreEquivvalent` on all 6 arrays to make sure they all share the same elements, but in different orders.

Good job overall.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -72,6 +72,22 @@ public void GetFilePath_ReturnsFileThatExists()
             Assert.AreEqual(expectedString, percentage);
         }
 
+        [TestMethod]
+        public void RandomizeQuestions_GivenValidArray_ReturnsRandomizedArray()
+        {
+            //Arrange
+            Question[] questions = new Question[3];
+            Question[] questionsCpy = new Question[questions.Length];
```
mmwoodfo, line 80:
```suggestion
            Question[] questionsCopy = new Question[questions.Length];
```

Not sure if this is a typo, but definitely make sure to spell out words rather than abbreviate in the majority of cases.

mmwoodfo (commented):
👍🏻

PR #20, opened 2022-01-15
Assignment 1 documentation update, pull request features complete.

SkylerJohnson102020:


tomclark-mtcsw:
My PR for Skyler's assignment is this:  "great minds think alike!"  LOL.  Your solution was pretty much the same as mine.  No issues that I can see.  Nice job.

SkylerJohnson102020:
@tomclark-mtcsw Whoa, I'll say, pretty close! Must have read the same StackOverflow! 😂

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -26,6 +26,25 @@ public void LoadQuestions_RetrievesQuestionsFromFile()
...
+        {
+            string filePath = Path.GetRandomFileName();
+            try
+            {
+                GenerateQuestionsFile(filePath, 2);
+
+                Question[] questions1 = Program.LoadQuestions(filePath);
+                Question[] questions2 = Program.LoadQuestions(filePath);
+
+                Assert.AreNotEqual(questions1, questions2);
```
quattro004, line 40: There's also a CollectionAssert object you can use which might be more appropriate. I don't think this assertion will compare the items if that's what you intended. https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.collectionassert.arenotequal?view=visualstudiosdk-2022

SkylerJohnson102020: The tests passed each time I ran it. Maybe I wasn't looking close enough.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -37,7 +37,7 @@ public void LoadQuestions_GenerateQuestionsInRandomOrder_Success()
                 Question[] questions1 = Program.LoadQuestions(filePath);
                 Question[] questions2 = Program.LoadQuestions(filePath);
 
-                Assert.AreNotEqual(questions1, questions2);
+                CollectionAssert.AreNotEqual(questions1, questions2);
```
SkylerJohnson102020, line 40: @quattro004 @Keboo According to what I have read, this should be a successful test since equality is determined by containing the same elements AND the same order. Since I am using Random, the order should be changed. However, I am still thinking the test is not bulletproof. I am opening and reading the Trivia.txt file and filtering it through Random randomizing the questions to ensure a different order each time, this is my intent. However, there is a small possibility, but it IS possible, that both instances will be randomized but be in the same order. What could the least hacky way to check this? Try/catch? A conditional checking if both instances are equal (same order) and if they are, re-instantiate questions2? Re-instantiate both? I am not 100% about this test.

SkylerJohnson102020: I guess one way to do it would be to instantiate question1 then run my randomize function in my test to produce questions 2, compare the two, then AssertNotEqual. That would save some time.

tomclark-mtcsw: I thought about this problem as well, mostly due to the small number of questions.  I'd feel more confident if the number of items to randomize was much larger.  For the purposes of this small assignment I decided that my initial test was sufficient.  But I do like the CollectionAssert.

SkylerJohnson102020: @tomclark-mtcsw Yeah, the test gives the right idea but little things like this bug me, pun intended. Glad I'm not the only one that thought about this flaw.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
mmwoodfo: yes, `Assert.AreNotEqual` will not do what you'd expect. This is asserting that the two objects are not the same, but no the collections. So for example, the following code would fail:

```C#
Question[] questions2 = new Question[questions1.Length];
questions1.CopyTo(questions2, 0);

Assert.AreEqual(questions1, questions2);
```

But, the following will pass:
```C#
Question[] questions2 = new Question[questions1.Length];
questions1.CopyTo(questions2, 0);

CollectionAssert.AreEqual(questions1, questions2);
```

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
mmwoodfo: Testing randomization is a difficult problem. I think one way to approach it is to increase your odds of it being random. One way to do this is to randomize a number of arrays, like 5 or 10 different arrays. Then you compare all of those arrays and assert that at least two of the indexes of all of the arrays differ. This would prove that the indexes move around and hence, prove randomization. This of course still has the potential to fail, but it's a decreased chance.

PR #30, opened 2022-01-17
Assignment1

mpennock5:
Submission of Assignment1 by @Aedyn01 and I.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -92,24 +92,31 @@ namespace PrincessBrideTrivia.Tests
         // Testing for the extra credit part of the assignment
         [TestMethod]
         [DataRow("a", "b", "c", "d", "e", "f", "g")]
-        public void GetRandomizedQuestions_ReturnsTrueIfEqual(params string[] questions) 
+        public void GetRandomizedQuestions_ReturnsTrueIfEqual(params string[] text) 
         {
             // Arrange
             List<string> originalQuestions = new List<string>();
             List<string> randomQuestions = new List<string>();
-            for (int i = 0; i < questions.Length - 1; i++)
+            Question[] questions = new Question[text.Length - 1];
+
+            for (int i = 0;i < text.Length -1; i++)
             {
-                originalQuestions.Add(questions[i]);
+                Question question = new Question();
+                question.Text = text[i];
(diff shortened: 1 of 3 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +26,8 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            double decimalPercent = (double)numberCorrectAnswers / numberOfQuestions;
+            return (int)(decimalPercent * 100) + "%";
```
mmwoodfo, line 30: Instead of casting back to an int, be more explicit about your intentions. Use Math.Round or another method for rounding your number.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -86,5 +88,40 @@ private static void GenerateQuestionsFile(string filePath, int numberOfQuestions
...
+            Program.Randomize(questions);
+
+            for (int i = 0; i < questions.Length - 1; i++)
+            {
+                randomQuestions.Add(questions[i].Text);
+            }
+
+            // Assert
+            bool isNotEqual = Enumerable.SequenceEqual(randomQuestions, originalQuestions);
+            Assert.IsFalse(isNotEqual);
```
mmwoodfo, line 124: You can also use CollectionAssert here to ensure equality (or rather, non-equality)

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -86,5 +88,40 @@ private static void GenerateQuestionsFile(string filePath, int numberOfQuestions
...
+                originalQuestions.Add(questions[i].Text);
+            }
+
+            // Act
+            Program.Randomize(questions);
+
+            for (int i = 0; i < questions.Length - 1; i++)
+            {
+                randomQuestions.Add(questions[i].Text);
+            }
```
mmwoodfo, line 120: Try to keep code under your `act` portion of the test as short as possible. This really should only be one line. A way to do this would be to set randomQuestions equal to the output of `Randomize` . Of course, you will have to account for the fact that questions is manipulated by your function, so additional re-work would be needed here.

PR #32, opened 2022-01-18
Assignment1

brucegomes:
- [x] Feature request issue 3
- [x] Feature request issue 4

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -1,5 +1,6 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System.IO;
+using System.Linq;
 
 namespace PrincessBrideTrivia.Tests
 {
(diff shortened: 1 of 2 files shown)


PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -86,5 +87,16 @@ private static void GenerateQuestionsFile(string filePath, int numberOfQuestions
...
+
+        [TestMethod]
+        public void AreQuestionsRandom()
+        {
+            string filePath = Program.GetFilePath();
+            Question[] questionSet1 = Program.LoadQuestions(filePath);
+            Question[] questionSet2 = Program.LoadQuestions(filePath);
+
+            // Assert
+            Assert.IsFalse(questionSet1.SequenceEqual(questionSet2));
```
mmwoodfo, line 99: You can also use `CollectionAssert` here.

mmwoodfo: There is a small chance that the randomization actually randomized the array back into the original order, which would cause this case to fail. A way around this is to increase the odds that it's randomized by either increasing the number of elements within the array or shuffling multiple arrays and ensuring that at least two indexes values across those arrays differ for a shared index.

PR #35, opened 2022-01-18
Assignment1

czamora7:
Fixed the crashing issue.


mmwoodfo (commented):
Good job on issue 1, but it doesn't look like you attempted issue 2.

PR #37, opened 2022-01-18
Assignment1 @Dakota Condos

htran9:


PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -6,6 +6,41 @@ namespace PrincessBrideTrivia.Tests
...
+                temp = Program.ShuffleQuestions(temp);
+                if (temp[0].Text.Equals("4") &&
+                    temp[1].Text.Equals("3") &&
+                    temp[2].Text.Equals("2") &&
+                    temp[3].Text.Equals("1"))
+                {
+                    loopFlag = true;
+                    conditionFlag = true;
+                }
+                if (counter > 500) { loopFlag = true; }
```
mmwoodfo, line 37: An interesting way to test this problem, but maybe not the best way. What you may want to consider doing instead is shuffling 5 arrays and making sure that at least two of the indexes across all 5 differ to show that the items have indeed moved around.

mmwoodfo (commented):
Looks good 👍🏻

PR #39, opened 2022-01-19
Assignment1 @Disciple4GIT @czamora7

nwnel-017:
System no longer crashes, unit tests all run.

### DIFF
--- a/ConsoleApp2/ConsoleApp2/Program.cs
+++ b/ConsoleApp2/ConsoleApp2/Program.cs
@@ -1,0 +1,2 @@
+﻿// See https://aka.ms/new-console-template for more information
+Console.WriteLine("Hello, World!");
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -24,7 +24,7 @@ namespace PrincessBrideTrivia
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (int)Math.Round((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
         }
 
         public static bool AskQuestion(Question question)
@@ -54,6 +54,8 @@ namespace PrincessBrideTrivia
 
         public static void DisplayQuestion(Question question)
         {
+       
+            //Exception thrown- object not set to instance of object question.Text
             Console.WriteLine("Question: " + question.Text);
             for (int i = 0; i < question.Answers.Length; i++)
             {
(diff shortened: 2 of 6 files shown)


ConsoleApp2/ConsoleApp2/Program.cs
```diff
@@ -0,0 +1,2 @@
+﻿// See https://aka.ms/new-console-template for more information
+Console.WriteLine("Hello, World!");
```
mmwoodfo, line 2: Be sure not to include files that are not part of the assignment in the future. These two console app solutions don't need to be in this PR.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+//Comment
```
mmwoodfo, line 3:
```suggestion

```

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +25,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (int)Math.Round((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
```
mmwoodfo, line 28: While this works, it does use some un-needed casting. While casting to an int isn't bad, it would be cleaner to be more explicit about your intentions here.

Instead of casting to an int here, you can use  `.ToString("N0")` 
See [documentation here](https://docs.microsoft.com/en-us/dotnet/api/system.double.tostring?view=net-6.0)

There are a number of other ways as well.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -54,6 +55,8 @@ public static bool DisplayResult(string userGuess, Question question)
 
         public static void DisplayQuestion(Question question)
         {
+       
+            //Exception thrown- object not set to instance of object question.Text
```
mmwoodfo, line 59: No need to add a comment

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -89,6 +92,7 @@ public static Question[] LoadQuestions(string filePath)
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
+                questions[i] = question; //Added by me
```
mmwoodfo, line 95: Who's 'me'? :)
No need for a comment here.

Good job with the solution!

mmwoodfo (commented):
Great job on the assignment guys. 

Please note, you are only supposed to pair with one other person for the assignment.

PR #43, opened 2022-01-20
Assignment2 @ Tony Chan

tnadalini:
initial PR (assignment not yet completed)


Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+
+                _FilePath = value;
+            }
+        }
+
+        public FileLogger(string? filePath) { FilePath = filePath; }  
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            string logResult = $"{DateTime.Now} {ClassName} {logLevel}: {message} \n";
```
mmwoodfo, line 25: String interpolation! The only suggestion I have here is to use `Environment.NewLine` instead of `\n`, but looks good otherwise. 
```suggestion
            string logResult = $"{DateTime.Now} {ClassName} {logLevel}: {message}{Environment.NewLine}";
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,44 @@
...
-            return null;
+            get => _FilePath!;
+
+            set
+            {
+                if (value == null) { throw new ArgumentNullException(nameof(value)); }
+
+                _FilePath = value;
+            }
+        }
```
mmwoodfo, line 18: Same comments I made for the `FileLogger.cs` apply here. (However, this can be nullable)

Logger/LogFactory.cs
```diff
@@ -1,11 +1,44 @@
...
+                    ClassName = className
+                }; 
+
+                return fileLogger;
+            }
+
+            return null; 
+        }
+
+        public void ConfigureFileLogger (string filePath)
```
mmwoodfo, line 38: 👍🏻

Logger/LogFactory.cs
```diff
@@ -1,11 +1,44 @@
...
+            set
+            {
+                if (value == null) { throw new ArgumentNullException(nameof(value)); }
+
+                _FilePath = value;
+            }
+        }
+
+
+        public BaseLogger? CreateLogger(string className, string? filePath)
```
mmwoodfo, line 21:
```suggestion
        public BaseLogger? CreateLogger(string className)
```

You should be using the filePath set from the `ConfigureFileLogger` method, not passing in a new filePath

Logger/LogFactory.cs
```diff
@@ -1,11 +1,44 @@
...
+        {
+            if (filePath != null)
+            {
+                ConfigureFileLogger(filePath);
+                
+                FileLogger fileLogger = new FileLogger(FilePath)
```
mmwoodfo, line 27: Notice here, you switch between using `filePath` and `FilePath`, even though you haven't verified that `FilePath` is not null. This can very easily lead to difficult-to-find bugs in your code.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,56 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        private static string? messageWithArgs;
```
mmwoodfo, line 7:
```suggestion

```

If you're going to use this as an intermediate variable between formatting and logging, you should make the scope of this variable local to the current function being called. 

Leaving it here, as a variable accessible to the scope of this entire class makes your intent less clear (confusing code) and leaves you open to possible bugs where a leftover string can be logged out instead of the string you intend.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,56 @@
...
+        public static void Error(this BaseLogger? logger, string message, params object[] args) 
+        {
+            if (logger == null)
+            {
+                throw new ArgumentNullException(nameof(logger));
+            }
+            
+            messageWithArgs = String.Format(message, args);
```
mmwoodfo, line 15:
```suggestion
            var messageWithArgs = string.Format(message, args);
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,56 @@
...
+        }
+
+        public static void Warning(this BaseLogger? logger, string message, params object[] args)
+        {
+            if (logger == null)
+            {
+                throw new ArgumentNullException(nameof(logger));
+            }
+
+            messageWithArgs = String.Format(message, args);
```
mmwoodfo, line 27:
```suggestion
            var messageWithArgs = string.Format(message, args);
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,56 @@
...
+        }
+
+        public static void Information(this BaseLogger? logger, string message, params object[] args)
+        {
+            if (logger == null)
+            {
+                throw new ArgumentNullException(nameof(logger));
+            }
+
+            messageWithArgs = String.Format(message, args);
```
mmwoodfo, line 39:
```suggestion
            var messageWithArgs = string.Format(message, args);
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,56 @@
...
+        }
+
+        public static void Debug(this BaseLogger? logger, string message, params object[] args)
+        {
+            if (logger == null)
+            {
+                throw new ArgumentNullException(nameof(logger));
+            }
+
+            messageWithArgs = String.Format(message, args);
```
mmwoodfo, line 51:
```suggestion
           var messageWithArgs = string.Format(message, args);
```

Logger/Logger.csproj
```diff
@@ -2,6 +2,12 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+	  <Nullable>enable</Nullable>
+	  <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	  <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
+	  <LangVersion>10.0</LangVersion>
```
mmwoodfo, line 9: Good job here, although you are supposed to enable these for both projects (Logger.Tests.csproj)

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,42 @@
...
 using System.IO;
+using System;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class LogFactoryTests
     {
+        LogFactory exLogFactory; //'example LogFactory
+        FileLogger exLogger; //'example Logger'
```
mmwoodfo, line 11: Again, keep these within the scopes of the individual tests
```suggestion

```

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,42 @@
...
+
+        [TestInitialize]
+        public void Initialize()
+        {
+            exLogFactory = new LogFactory();
+            exLogger = (FileLogger) exLogFactory.CreateLogger("fileLog", "C://Egg");
+        }
```
mmwoodfo, line 18: There shouldn't be any need for a test initializer, as you probably want to make a fresh class for each test in this case. Also, you should be using `nameof()` here to get the classname.

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,42 @@
...
+        {
+            FileLogger logger = (FileLogger) exLogFactory.CreateLogger("fileLog1", "C://Egg");
+
+            //Assert.Equals(exLogger.ClassName, logger.ClassName);
+            Assert.AreEqual(exLogger.FilePath, logger.FilePath);
+            Assert.AreEqual("fileLog1", logger.ClassName);
+        }
+
+        [TestMethod]
+        public void NullFilePath_ThrowNullException()
```
mmwoodfo, line 31:
```suggestion
        public void NullFilePath_ReturnsNull()
```

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -14,19 +14,60 @@ public void Error_WithNullLogger_ThrowsException()
...
+            // Act
+            BaseLoggerMixins.Warning(null, "");
+
+            // Assert
+            //Assert.ThrowsException<ArgumentNullException>();
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Debug_WithNullLogger_ThrowsException()
```
mmwoodfo, line 51: In addition to testing the null inputs, you should also test that it logs the message correctly when the input is not null.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -5,6 +5,30 @@ namespace Logger.Tests
     [TestClass]
     public class FileLoggerTests
     {
-        
+
+        LogFactory logFactory;
+        FileLogger fileLogger;
+
+        [TestInitialize]
+        public void TestInitialize()
```
mmwoodfo, line 13: You should have some tests in here that confirms the logger is appending to the end of a file, logging correct log levels, formatting correctly, etc.

PR #47, opened 2022-01-25
Assignment2

TechGlue:
Initial PR. 
Pair Code Partner: @NicholasJohnsonDev Nicholas Johnson


Ashrom0:
I will review this

Logger/FileLogger.cs
```diff
@@ -0,0 +1,47 @@
+using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string? ClassName { get; set; }
```
mmwoodfo, line 8: This should have been declared in `BaseLogger.cs` and inherited here.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,47 @@
...
+            FilePath = filePath;
+        }
+
+        public FileLogger(string? className, string? filePath, LogLevel status, string? message)
+        {
+            ClassName = className;
+            FilePath = filePath;
+            Status = status;
+            Message = message;
+        }
```
mmwoodfo, line 25: You should only have one constructor here that simply takes the file path. The rest of these parameters should not be class level properties, but rather scoped to only the log method.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public string ToText()
```
mmwoodfo, line 27: If you're going to separate this out, you may consider making it private as you only want people calling the `Log` method from this class.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            return $"{DateTime.Now:MM/dd/yyyy hh:mm:ss tt}\n{ClassName}\n{Status}\n{Message}\n";
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            Status = logLevel;
+            Message = message;
+            string filePath = "";
```
mmwoodfo, line 36: Why not just use `FilePath`? There isn't any need to create another variable for it here.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,47 @@
...
+            if (FilePath != null)
+            {
+                filePath = FilePath;
+            }
+            var fs = new FileStream(filePath, FileMode.Append, FileAccess.Write);
```
mmwoodfo, line 41: Consider using a `using` here for your `FileStream` to automatically clean it up at the end of it's lifetime. Here is some documentation on it: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement

Logger/FileLogger.cs
```diff
@@ -0,0 +1,47 @@
...
+            StreamWriter sw = new(fs);
```
mmwoodfo, line 42: Same thing here with the StreamWriter, `using` would be good. Also, you could just use the `StreamWriter`, no need for both.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,47 @@
...
+            sw.WriteLine(ToText());
```
mmwoodfo, line 43: Forgetting to close the FileStream?

Logger/LogFactory.cs
```diff
@@ -1,11 +1,16 @@
...
-            
-            return null;
-        }
-    }
-}
+﻿namespace Logger
+{
+    public class LogFactory
+    {
+        public string? FilePath { get; private set; }
```
mmwoodfo, line 5: This looks good! 👍🏻 Nice private set.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,55 @@
...
-    }
-}
+﻿using System;
+using System.Globalization;
+
+namespace Logger
+{
+    public static class BaseLoggerMixins
+    {
+        public static void Error(string? message, BaseLogger logger, params object[] arguments)
```
mmwoodfo, line 8:
```suggestion
        public static void Error(string? message, this BaseLogger logger, params object[] arguments)

```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,55 @@
...
+        {
+            if (logger is null)
+                throw new ArgumentNullException(nameof(logger));
+
+            if (message != null)
```
mmwoodfo, line 13: You could also make it so `message` isn't nullable as a parameter.

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,42 @@
...
+    public class LogFactoryTests
+    {
+
+        [TestMethod]
+        public void ConfigureFileLogger_WithValidFilePath_PrivateMemberFilePathIsSet()
+        {
+            //Arrange
+            LogFactory logFactory = new();
+            //Act
+            logFactory.ConfigureFileLogger("/Users/luis/");
```
mmwoodfo, line 13: You may consider storing this string in a variable, so you don't have to type it twice.

mmwoodfo:
@TechGlue Did you do a PR review for Assignment 2? I can't seem to find one.

PR #48, opened 2022-01-25
Assignment2

tomclark-mtcsw:


Logger.Tests/LogFactoryTests.cs
```diff
@@ -6,6 +6,31 @@ namespace Logger.Tests
...
     {
-        
+        [TestMethod]
+        public void CreateLogger_WithEmptyFilePath_ReturnsNull()
+        {
+            // Arrange
+            var logFactory = new LogFactory();
+
+            // Act
+            var logger = logFactory.CreateLogger(nameof(LogFactoryTests));
```
whitneybolar, line 16: Oh, I did not think to use nameof() in the tests. My partner and I were confused by this, but that makes total sense.

Logger.Tests/Logger.Tests.csproj
```diff
@@ -3,6 +3,18 @@
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
     <IsPackable>false</IsPackable>
+	<Nullable>enable</Nullable>
+	<EnforceCodeStyleInBuild>False</EnforceCodeStyleInBuild>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
whitneybolar, line 12: I did not to this on my project. Is there any major changes to your output when you use this?

Logger/BaseLogger.cs
```diff
@@ -2,6 +2,12 @@
 {
     public abstract class BaseLogger
     {
+        public string ClassName { get; set; } = string.Empty;
+
+        public BaseLogger()
```
whitneybolar, line 7: Is the constructor being used? If not, do you think you should like to take it out?

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,75 @@
...
+        public static void Error(this BaseLogger? baseLogger, string message, params object[] parameters)
+        {
+            if (baseLogger == null)
+            {
+                throw new ArgumentNullException("BaseLogger is null!");
+            }
+            else
+            {
+                string logMessage = message;
+                if (parameters.Length > 0)
```
whitneybolar, line 16: I did not think to put in that condition. That is a good idea. That way we do not need to waste carrying parameters if its empty.

whitneybolar (commented):
There is an existing BaseLogger class. It needs an auto property to hold the class name. This property should be set in the LogFactory using an object initializer. ✔ it is an empty string 
The current date/time ✔
The log level ✔
The message ✔
The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message"
If the file logger has not be configured in the LogFactory, its CreateLogger method should return null. ✔ cannot be configured if the filePath is null or empty, nice!
Error, ✔
Warning, ✔
Information, and ✔
Use the nameof() operator when identifying the class name to the logger ✔ in test
Set LangVersion to 10.0 ❌✔ not in Program.csproj or Logger.csproj, quick fix!
Extra Credit
Implement an additional logger. This logger must be unit tested. Some options to consider could be one that uses System.Console or System.Diagnostics.Trace ❌ I see one logger test

Logger.Tests/Logger.Tests.csproj
tomclark-mtcsw: None that I saw.  Thanks for the note about the language version.  It's set automatically in Visual Studio but I know they want an explicit tag in the csproj file.

Logger/BaseLogger.cs
tomclark-mtcsw: You are correct.  Because this is an abstract class, the constructor is not necessary.  Good catch!

Logger/FileLogger.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        public string LogFilePath { get; set; } = string.Empty;
```
mmwoodfo, line 10:
```suggestion
        public string LogFilePath { get; set; }
```

If you remove your default constructor on lines 12-15, you should be able to do the above and avoid a nullable error.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                UpdateLog(logLevel, message, LogFilePath);
+            }
+            else
+            {
+                throw new ArgumentException("LogFilePath can not be empty!");
+            }
+        }
+
+        public virtual void UpdateLog(LogLevel logLevel, string message, string filePath)
```
mmwoodfo, line 33: Make this private perhaps?

Logger/LogFactory.cs
```diff
@@ -2,10 +2,27 @@
 {
     public class LogFactory
     {
-        public BaseLogger CreateLogger(string className)
+        private string FilePath = "";
```
mmwoodfo, line 5:
```suggestion
        private string? _FilePath;
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,27 @@
...
+            FileLogger fileLogger = new FileLogger()
+            {
+                ClassName = className
+            };
+
+            if (!string.IsNullOrEmpty(FilePath))
+            {
+                fileLogger.LogFilePath = FilePath;
+                return fileLogger;
+            }
```
mmwoodfo, line 18:
```suggestion
           if(string.IsNullOrWhiteSpace(FilePath) return null;
           
           return new FileLogger(FilePath) { ClassName = className };
```

PR #49, opened 2022-01-25
Assignment2

sammyboy2000:
The initial pull request.

### DIFF
--- a/Logger.Tests/BaseLoggerMixinsTests.cs
+++ b/Logger.Tests/BaseLoggerMixinsTests.cs
@@ -23,7 +23,7 @@ namespace Logger.Tests
         public void Error_WithData_LogsMessage()
         {
             // Arrange
-            TestLogger? logger = new TestLogger();
+            TestLogger? logger = new();
 
             // Act
             logger.Error("Message {0}", 42);
--- a/Logger.Tests/BaseLoggerMixinsTests.cs
+++ b/Logger.Tests/BaseLoggerMixinsTests.cs
@@ -14,7 +14,7 @@ namespace Logger.Tests
             // Arrange
 
             // Act
-            //BaseLoggerMixins.Error(null, "");
+            BaseLoggerMixins.Error(null!, "");
 
             // Assert
         }
@@ -23,10 +23,10 @@ namespace Logger.Tests
         public void Error_WithData_LogsMessage()
         {
             // Arrange
-            var logger = new TestLogger();
+            TestLogger? logger = new TestLogger();
 
             // Act
-            //logger.Error("Message {0}", 42);
+            logger.Error("Message {0}", 42);
 
             // Assert
             Assert.AreEqual(1, logger.LoggedMessages.Count);
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -5,6 +5,7 @@ namespace Logger.Tests
     [TestClass]
     public class FileLoggerTests
     {
-        
+
     }
+
 }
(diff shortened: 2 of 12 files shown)


GavenElder:
I will review this.

GavenElder:
There is an existing BaseLogger class. It needs an auto property to hold the class name. This property should be set in the LogFactory using an object initializer. ✔
        The current date/time ✔
        The log level ✔
        The message ✔
        The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message"
    If the file logger has not be configured in the LogFactory, its CreateLogger method should return null. ✔
        Error, ✔
        Warning, ✔
        Information, and ✔
    Use the nameof() operator when identifying the class name to the logger ✔
    Set LangVersion to 10.0 ✔

Extra Credit

    Implement an additional logger. This logger must be unit tested. Some options to consider could be one that uses System.Console or System.Diagnostics.Trace ❌

GavenElder:
Overall this code works as intended. It didn't crash on my end. Seems to check all boxes with the assignment checklist, besides for the extra credit section. Great work.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        public string? FilePath { get; set; }
+        //string? ClassName inherited from BaseLogger
+        public FileLogger(string className, string filePath)
```
mmwoodfo, line 10: Not required to take in the class name here, but doesn't hurt.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            ClassName = className;
+            FilePath = filePath;
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            //example: 10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message
+            File.AppendAllText(FilePath, $"{DateTime.Now} {ClassName} {logLevel}: {message} \n");
```
mmwoodfo, line 19:
```suggestion
            File.AppendAllText(FilePath, $"{DateTime.Now} {ClassName} {logLevel}: {message}{Environment.NewLine}");
```
Looks great, my only suggestion would be to use `Environment.NewLine` instead of `\n`

Logger/LogFactory.cs
```diff
@@ -1,11 +1,42 @@
...
+        }
+
+        public static BaseLogger ConfigureConsoleLogger(string className)
+        {
+            return new ConsoleLogger(className);
+        }
+
+        public static BaseLogger ConfigureFileLogger(string className, string? filePath)
+        {
+            if (String.IsNullOrEmpty(filePath))
```
mmwoodfo, line 33: This is technically allowed by the assignment instructions.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,33 @@
...
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        private readonly string ClassName = "FileLoggerTests";
+
+        [TestMethod]
+        public void FileLogger_GivenError_Log()
```
mmwoodfo, line 12: This test doesn't seem to check that the log was in fact an error log, just that the log was written to the file. There should be some more tests here.

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -23,10 +23,10 @@ public void Error_WithNullLogger_ThrowsException()
         public void Error_WithData_LogsMessage()
         {
             // Arrange
-            var logger = new TestLogger();
+            TestLogger? logger = new();
```
mmwoodfo, line 26: More tests would be beneficial in this file too.

PR #54, opened 2022-01-25
Assignment2

czamora7:


Ashrom0 (commented):
There is an existing BaseLogger class. It needs an auto property to hold the class name. This property should be set in the LogFactory using an object initializer. ✔
The current date/time ✔
The log level ✔
The message ✔
If the file logger has not be configured in the LogFactory, its CreateLogger method should return null. ✔
Error, ✔
Warning, ✔
Information, and ✔
Use the nameof() operator when identifying the class name to the logger ✔
Set LangVersion to 10.0 ✔

Overall seems pretty good

PR #61, opened 2022-01-27
Assignment2 @Hieu Tran

DakotaCondos:
Dakota Condos
Hieu Tran
Attempted extra credit using a ConsoleLogger


Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+    public string? LogText => _LogText;
+    public FileLogger(string className, string path)
+    {
+        ClassName = className;
+        filePath = path;
+    }
+    public override void Log(LogLevel logLevel, string message)
+    {
+        StreamWriter sw = File.AppendText(filePath);
+        _LogText = $"{DateTime.Now} {nameof(FileLogger)} {logLevel} {message}";
```
mmwoodfo, line 18:
```suggestion
        _LogText = $"{DateTime.Now} {ClassName} {logLevel} {message}";
```
`nameof` should be used when setting the classname.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+    private string? _LogText;
+    public string? LogText => _LogText;
+    public FileLogger(string className, string path)
+    {
+        ClassName = className;
+        filePath = path;
+    }
+    public override void Log(LogLevel logLevel, string message)
+    {
+        StreamWriter sw = File.AppendText(filePath);
```
mmwoodfo, line 17: Consider using [`using`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement) here.

Logger/LogFactory.cs
```diff
@@ -1,11 +1,33 @@
...
+            }
+
+
+            return className == "FileLogger" ? new FileLogger(className, FilePath) : new ConsoleLogger();
+        }
+
+        public void ConfigureFileLogger(string filePath)
+        {
+            _FilePath = filePath;
+            IsLogConfigured = true;
```
mmwoodfo, line 30: This boolean isn't needed here, simply make `_FilePath` nullable and check if it's null or not to determine if it has been configured.

PR #64, opened 2022-01-28
Assignment2

schuywalker:
Partnered with @whitneybolar


Logger/BaseLogger.cs
```diff
@@ -1,7 +1,20 @@
...
     }
+
 }
+
+
+
+
+
+
+
```
mmwoodfo, line 20: Logs of whitespace

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,20 @@
...
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace Logger { 
+
     public abstract class BaseLogger
     {
-        public abstract void Log(LogLevel logLevel, string message);
+        public abstract void Log(LogLevel? logLevel, string message);
+
+        public string? ClassName { get; set; } = string.Empty;
```
mmwoodfo, line 9:
```suggestion
        public string? ClassName { get; set; }
```
No need to set this to an empty string. It can be left as null.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,46 @@
...
+{ 
+    //References for research
+    //https://stackoverflow.com/questions/20185015/how-to-write-log-file-in-c
+    //https://docs.microsoft.com/en-us/dotnet/standard/io/how-to-open-and-append-to-a-log-file
+
+    public class FileLogger : BaseLogger
+    {
+
+        public string FilePath 
+        { get; }
```
mmwoodfo, line 18: Keep this one one line.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,46 @@
...
+            //https://www.c-sharpcorner.com/UploadFile/mahesh/create-a-text-file-in-C-Sharp/
+            
+
+            // * Code Reference: https://docs.microsoft.com/en-us/dotnet/standard/io/how-to-open-and-append-to-a-log-file
+            using (StreamWriter w = File.AppendText("file.txt"))
+            {
+                w.Write(dateTime + "\n");
+                w.Write(this.ClassName + "\n");
+                w.Write(logLevel + "\n");
+                w.Write(message);
```
mmwoodfo, line 40: This log should be logged on a single line.

Logger/LogFactory.cs
```diff
@@ -1,11 +1,42 @@
...
+                };  
+                return logger;
+            }
+               
+        }
+
+        public void ConfigureFileLogger(string FilePath)
+        {
+                Path = FilePath;
+                Configured = true;
```
mmwoodfo, line 38: No need to use a boolean here. Just check if `Path` is null or not to determine if it's been configured.

Logger/LogFactory.cs
```diff
@@ -1,11 +1,42 @@
...
+                    ClassName = className,
+                };  
+                return logger;
+            }
+               
+        }
+
+        public void ConfigureFileLogger(string FilePath)
+        {
+                Path = FilePath;
```
mmwoodfo, line 37:
```suggestion
        public void ConfigureFileLogger(string filePath)
        {
                Path = flePath;
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,42 @@
...
+        public BaseLogger? CreateLogger(string className) //BaseLogger
+        {
+            if (!Configured)
+            {
+                return null!;//intending to return null in this case   
+            }
+
+            else
+            {
+                BaseLogger logger = new FileLogger(Path!) //using bang because if condition will execute above if Path is Null
```
mmwoodfo, line 26: This is why you should check if the file is null or not, rather than checking a condition. Then you won't have to use the `!` operator.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,92 @@
...
+
+        // * Code Reference
+        // https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger?view=dotnet-plat-ext-3.1#extension-methods
+        public static void LogError(this BaseLogger? baseLogger, string message, params object[] args)
+        {
+            if (baseLogger is null)
+            {
+                throw new ArgumentNullException(nameof(baseLogger));
+            }
+            if (message is null)
```
mmwoodfo, line 19: Shouldn't need to check if the message is null here.

mmwoodfo (commented):
Good job

PR #70, opened 2022-02-01
Assignment3

Ashrom0:


czamora7 (commented):

Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ❌

Fundamentals
Ensure you enable:
net6 targetted ✔
C# 10.0 ✔
All of the above should be unit tested ❌

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,68 @@
...
+        [TestMethod]
+        public void TellJoke_ChuckNorrisTest_FindNewJoke()
+        {
+
+            Mock<JokeService> mock = new Mock<JokeService>();
+
+            mock.SetupSequence(jokeService => jokeService.GetJoke())
+                .Returns("Chuck Norris does not sleep. He waits.")
+                .Returns("Chuck Norris can divide by zero, and counted to infinity 3 times.")
+                .Returns("When Chuck Norris falls in water, Chuck Norris doesn't get wet. Water gets Chuck Norris.")
```
mmwoodfo, line 56:
```suggestion
            mock.SetupSequence(jokeService => jokeService.GetJoke())
                .Returns("Chuck Norris does not sleep. He waits.")
                .Returns("Chuck Norris can divide by zero, and counted to infinity 3 times.")
                .Returns("When Chuck Norris falls in water, Chuck Norris doesn't get wet. Water gets Chuck Norris.");
```

mmwoodfo: This is causing your code not to compile.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,68 @@
...
+
+        [TestMethod]
+        public void Jester_WorkingJokeOutput()
+        {
+            JokeService jokeService = new JokeService();
+            JokeOutput jokeOutput = new JokeOutput();
+
+            Jester jester = new Jester(jokeService, jokeOutput);
+
+            Assert.AreEqual(jokeService, jokeOutput);
```
mmwoodfo, line 32: Why check if these are equal? Shouldn't you check that this doesn't throw an exception if a null input is introduced?

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+
+namespace CanHazFunny
+{
+    public class Jester
+    {
+        private JokeService jokeService;
+        private JokeOutput jokeOutput;
```
mmwoodfo, line 8:
```suggestion
        private JokeService JokeService {get;}
        private JokeOutput JokeOutput {get;}
```

PR #71, opened 2022-02-01
Assignment3

sammyboy2000:


Ashrom0:
I will review this.

Ashrom0 (commented):
The JokeService will need to have the interface applied to it. ❌
The Jester class TellJoke() method should retrieve a joke from the JokeService. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ❌
Ensure you enable:
net6 targetted ✔
C# 10.0 ✔

Ashrom0:
Although the program seems to run good. I see that the JokeService was taken out and replaced for another method and since its missing, I dont see the telljoke method that retrieves from jokeservices, but you still were able to run the program just fine.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Net.Http;
+
+namespace CanHazFunny
+{
+    public class Jester : IOutput
+    {
+        public HttpClient HttpClient { get; }
```
mmwoodfo, line 8: This should be the IOutput class.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,40 @@
...
+        public JokeService JokeService { get; }
+        public Jester()
+        {
+            HttpClient = new HttpClient();
+            JokeService = new JokeService();
+        }
```
mmwoodfo, line 14:
```suggestion
        public Jester(IJokeService jokeService, IOutputService outputService)
        {
            OutputService = outputService ?? null check here;
            JokeService = jokeService ?? null check here;
        }
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,40 @@
...
+            string joke;
+            do
+            {
+                joke = JokeService.GetJoke(HttpClient);
+            } while (!ScreenJoke(joke, filters));
+
+            IOutput.PresentJoke(joke);
+        }
+
+        public static bool ScreenJoke(string joke, string[] filters)
```
mmwoodfo, line 27: Maybe make this private?

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            if (joke == null || filters == null || filters.Length == 0)
```
mmwoodfo, line 29: Good idea making the filters a passed-in a variable. It makes your method a lot more flexible and easy to change/refactor in the future.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,60 @@
...
+using System;
+using System.IO;
 
 namespace CanHazFunny.Tests
 {
     [TestClass]
     public class JesterTests
     {
-        
+        [TestMethod]
```
mmwoodfo, line 10: You could benefit from more tests here, including tests for the JokeService class.

PR #73, opened 2022-02-01
Assignment3

tomclark-mtcsw:


schuywalker:
**Beautiful code! I don't fully understand the Unicode Converter but it looks like it's working. I can't come up with any suggestions.**

Fundamentals
Ensure you enable:
net6 targetted ✔
C# 10.0 ✔
Extra Credit
Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)?
The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string.
Additional links
Moq for test doubles
coverlet code coverage tool
Geek jokes API

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+        else
+        {
+            OutputText(theJoke);
+            success = true;
+        }
+        return success;
+    }
+
+    public static bool IsChuckNorrisJoke(string joke)
```
mmwoodfo, line 46: Maybe make this private? But it doesn't matter that much.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace CanHazFunny;
+
+public class Jester : IJokeService, IJokeOutput
```
mmwoodfo, line 7: I believe instead of inheriting, the assignment wanted you to implement these two interfaces in seperate classes and take those classes as dependencies in the `Jester` constructor.

CanHazFunny/CanHazFunny/JokeModel.cs
```diff
@@ -0,0 +1,10 @@
+﻿
+namespace CanHazFunny
+{
+    public class JokeModel
+    {
+        // Geek jokes API model uses lowercase in JSON
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "INTL0002:Properties PascalCase", Justification = "<Pending>")]
+        public string joke { get; set; } = string.Empty;
```
mmwoodfo, line 8:
```suggestion
        [JsonProperty("joke")]
        public string Joke { get; set; } = string.Empty;
```

You could use the `JsonProperty` decorator to specify the json name for this property. (You might need to import `Newtonsoft.json` to use this)

PR #76, opened 2022-02-01
Assignment3

schuywalker:
still lots of work to do on tests. Currently stuck on why GetJoke_ReturnsNonNullString_Success() in JokeServiceTests isnt passing/ runs indefinitely. If it returns a string, why can't we check if that string isnt null?

### DIFF
--- a/CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
@@ -28,9 +28,9 @@ namespace CanHazFunny.Tests;
         {
         ConsoleOutput consoleOutput = new();
         Assert.ThrowsException<ArgumentNullException>(() => consoleOutput.WriteJokeToConsole(null!));
-            // ! because it is intended to be null. (warning says cannot convert null to non-nullable reference type)
+           
         }
-        [TestMethod] // HAS TO BE RUN IN DEBUG MODE - NEED TO FIX THIS
+        [TestMethod]
         public void WriteJokeToConsole_GoodArgument_Success()
         {
             ConsoleOutput consoleOutput1 = new();
--- a/CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
@@ -27,7 +27,7 @@ namespace CanHazFunny.Tests;
         public void WriteJokeToConsole_NullArgument_Exception()
         {
         ConsoleOutput consoleOutput = new();
-        Assert.ThrowsException<ArgumentNullException>(() => consoleOutput.writeJokeToConsole(null!));
+        Assert.ThrowsException<ArgumentNullException>(() => consoleOutput.WriteJokeToConsole(null!));
             // ! because it is intended to be null. (warning says cannot convert null to non-nullable reference type)
         }
         [TestMethod] // HAS TO BE RUN IN DEBUG MODE - NEED TO FIX THIS
@@ -36,7 +36,7 @@ namespace CanHazFunny.Tests;
             ConsoleOutput consoleOutput1 = new();
             string joke = "test output string";
             
-            consoleOutput1.writeJokeToConsole(joke);
+            consoleOutput1.WriteJokeToConsole(joke);
             string? whatsInConsole = Console.ReadLine();
             
             Assert.IsNotNull(whatsInConsole);
--- a/CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
@@ -10,27 +10,18 @@ namespace CanHazFunny.Tests;
     [TestClass]
     public class ConsoleOutputTests
     {
-     /*   
-        [TestInitialize]
-        public void testInit()
-        {
-            ConsoleOutput consoleOutput = new();
-        }
-        [TestCleanup]
-        public void testCleanUp()
-    {
-            consoleOutput = null;
-    }
-        
-        */
+
         [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
         public void WriteJokeToConsole_NullArgument_Exception()
         {
         ConsoleOutput consoleOutput = new();
-        Assert.ThrowsException<ArgumentNullException>(() => consoleOutput.WriteJokeToConsole(null!));
+        consoleOutput.WriteJokeToConsole(null!);
+       // Assert.ThrowsException<ArgumentNullException>(() => consoleOutput.WriteJokeToConsole(null!));
            
         }
-        [TestMethod]
+       /* [TestMethod] // method times out and then says it's been run when I hit stop. Not sure why, but it's an unnecessary test.
+        *leaving in her until I understand what's happening with it.
         public void WriteJokeToConsole_GoodArgument_Success()
         {
             ConsoleOutput consoleOutput1 = new();
@@ -41,7 +32,7 @@ namespace CanHazFunny.Tests;
             
             Assert.IsNotNull(whatsInConsole);
             Assert.AreEqual(joke, whatsInConsole);
-        }
+        }*/
 
     }
     
--- a/CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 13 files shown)


tremorris1999:
I will review

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,50 @@
...
+        
+        private IJokeService? jokeService { get; set; } // set in constructor
+        private IConsoleOutput? consoleOutput { get; set; }
+
+        public Jester() // for mocks...
+        {
+
+        }
+
+         public Jester(IJokeService IjokeService, IConsoleOutput IconsoleOutput)
```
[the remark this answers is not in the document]
tremorris1999: I would also recommend you change the parameter names (e.g. jokeService and consoleOutput) to ensure proper camelCase convention.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+    public class Jester 
+    {   
+        
+        private IJokeService? jokeService { get; set; } // set in constructor
```
tremorris1999, line 12: No need to make these nullable since you're checking for null values in the constructor.
```suggestion
        private IJokeService jokeService { get; set; } // set in constructor
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+            jokeService = IjokeService;
+            consoleOutput = IconsoleOutput;
+         }
+       
+       
+       
+        public void TellJoke()
+        {
+            string joke = jokeService!.GetJoke(); // ! because TellJoke cant be called without instance, and constructor checks for null
```
tremorris1999, line 34: Implementing my above suggestions will eliminate the need to tell the compiler jokeService is not null.
```suggestion
            string joke = jokeService.GetJoke();
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,50 @@
...
+                joke = jokeService.GetJoke();
+                
+                failSafeCounter++;
+                if (failSafeCounter > 20)
+                {
+                    Console.WriteLine("Either 20 chuck norris jokes in a row or there is a bug");
+                    break;
+                }
+            }
+            consoleOutput!.WriteJokeToConsole(joke); // same as above - ! because TellJoke cant be called without instance, and constructor checks for null
```
tremorris1999, line 47: Implementing my above suggestions will eliminate the need to tell the compiler consoleOutput is not null.
```suggestion
            consoleOutput.WriteJokeToConsole(joke);
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,50 @@
...
+            jokeService = IjokeService;
+            consoleOutput = IconsoleOutput;
+         }
+       
+       
+       
+        public void TellJoke()
+        {
+            string joke = jokeService!.GetJoke(); // ! because TellJoke cant be called without instance, and constructor checks for null
+            int failSafeCounter = 0;
```
tremorris1999, line 35: Not sure what any of the failSafe business is? It shouldn't be needed.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,50 @@
...
+            while (joke.Contains("Chuck Norris"))
```
tremorris1999, line 36: I would recommend finding a way to compare these without case sensitivity. Not sure if it would be necessary since we don't know the joke API.

tremorris1999 (commented):
Hey there! Sorry if anything here came off as rude or pretentious! Tone is hard to convey over text. Overall, it seems to be in good function. I tried to be as thorough as I could, but I'm sure I missed a few things. Generally speaking, try and be more aware (or intentional) of when things are (or can be) null, as well as when non-nullability is in question too. I think this is causing most of your issues!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,35 @@
...
     {
-        
+        Jester? jester;
+
+        [TestMethod]
+        public void Jester_NullArguments_Exception()
+        {
+            Jester jester;
+            ConsoleOutput consoleOutput = new();
+            JokeService jokeService = null!;
```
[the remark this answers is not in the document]
tremorris1999: Solution in my comments on _Jester.cs_ (https://github.com/IntelliTect-Samples/EWU-CSCD371-2022-Winter/pull/76#discussion_r797440050)

CanHazFunny/CanHazFunny/Jester.cs
schuywalker: good idea!

CanHazFunny/CanHazFunny/Jester.cs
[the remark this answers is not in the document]
schuywalker: that was the issue I was stuck on - thank you!

schuywalker:
@tremorris1999 
thank you for reviewing my code so thoroughly!! I know mine wasn't the easiest to review since it still had so many kinks to iron out, but I so appreciate your time and comments! 
Tuesday's lecture gave me a much better grasp off the ? and ! operators which, as you mentioned, were the source of a lot of my troubles. Thanks again :)

CanHazFunny/CanHazFunny/Jester.cs
schuywalker: was just trying this because my test was running indefinitely

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,46 @@
...
+    private IConsoleOutput ConsoleOutput { get; set; }
+
+
+
+    public Jester(IJokeService IjokeService, IConsoleOutput IconsoleOutput)
+    {
+        if (IconsoleOutput is null || IjokeService is null)
+        {
+            throw new ArgumentNullException("an arg was null");
+        }
```
mmwoodfo, line 22: I would suggest checking each one individually and throwing an exception that lets the user know exactly what is null, rather than checking both at once and throwing a generic exception message like this.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public class Jester
+{
+
+    private IJokeService JokeService { get; set; } // set in constructor
+    private IConsoleOutput ConsoleOutput { get; set; }
```
mmwoodfo, line 13: You could make these readonly

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,46 @@
...
+
+
+
+    public Jester(IJokeService IjokeService, IConsoleOutput IconsoleOutput)
```
mmwoodfo, line 17:
```suggestion
    public Jester(IJokeService jokeService, IConsoleOutput consoleOutput)
```
While you are typing the parameters as their interface types, the values actually being passed will not be the interface, but rather an instance of a class that implements that interface. So I would suggest naming your input variables without the `i`

CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+       /* [TestMethod] // method times out and then says it's been run when I hit stop. Not sure why, but it's an unnecessary test.
+        *leaving in her until I understand what's happening with it.
+        public void WriteJokeToConsole_GoodArgument_Success()
+        {
+            ConsoleOutput consoleOutput1 = new();
+            string joke = "test output string";
+            
+            consoleOutput1.WriteJokeToConsole(joke);
+            string? whatsInConsole = Console.ReadLine();
```
mmwoodfo, line 31: This is the reason your test hangs and then times out. It is waiting for user input.

CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace CanHazFunny.Tests;
+
+    [TestClass]
+    public class ConsoleOutputTests
```
mmwoodfo, line 11: I am going to give you a point for the attempt at the extra credit.

PR #78, opened 2022-02-01
Assignment3 @ Tony Chan

tnadalini:


CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,30 @@
...
+        
+        public void TellJoke()
+        {
+            do
+            {
+                _joke = _jokeService.GetJoke();
+            }
+            while (_joke.Contains("Chuck Norris", StringComparison.OrdinalIgnoreCase));
+
+            _printService!.Print(_joke);
```
mmwoodfo, line 26: If you make this a non-nullable class field then you won't need to use `!`

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+namespace CanHazFunny
+{
+    public class Jester
+    {
+        private JokeService _jokeService;
+        private PrintService? _printService;
+        private string? _joke;
```
mmwoodfo, line 9: You should remove this as a class field and just use it as a local variable in the method it's scoped to.

PR #79, opened 2022-02-01
Assignment3

A-Tanner:
Extra Credit Attempted. JSON is handled using JsonJokeService, which is derived from JokeService.

### DIFF
--- a/CanHazFunny/CanHazFunny/CanHazFunny.csproj
+++ b/CanHazFunny/CanHazFunny/CanHazFunny.csproj
@@ -3,6 +3,11 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
+	  <Nullable>enable</Nullable>
+	  <LangVersion>10.0</LangVersion>
+	  <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
+	  <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
   </PropertyGroup>
 
 </Project>
--- a/CanHazFunny/CanHazFunny/JokeService.cs
+++ b/CanHazFunny/CanHazFunny/JokeService.cs
@@ -2,14 +2,21 @@
 
 namespace CanHazFunny
 {
-    public class JokeService
+    public class JokeService :IServiceable
     {
+
+        private string? URL { get; set; }
         private HttpClient HttpClient { get; } = new();
 
         public string GetJoke()
         {
-            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
+            string joke = HttpClient.GetStringAsync(URL).Result;
             return joke;
+        }
+
+        public void GetService(string serviceAcsess)
+        {
+            URL = serviceAcsess;
         }
     }
 }
--- a/CanHazFunny/CanHazFunny/JokeService.cs
+++ b/CanHazFunny/CanHazFunny/JokeService.cs
@@ -6,7 +6,7 @@ namespace CanHazFunny
     {
 
         public string? URL { get; private set; }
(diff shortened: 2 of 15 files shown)


GavenElder:
I will PR this.

GavenElder:
For this assignment do the following.

Fundamentals

    Ensure you enable:
        net6 targetted ✔
        C# 10.0 ✔

    (I didn't see the use Assert.AreEqual<T>() within assignment.)

   (besides for the exception of the "coverage.cobertura.xml" file) This one is rather lengthy and complicated.)

Extra Credit

    Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)?
    The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string.  ❌

Doesn't seem like the extra credit portion is completely done. But has been attempted.

A-Tanner:
Hey Gavin! Thanks for the peer review.

I had a couple of questions ;

What exception are you getting for coverage.cobertura.xml? This is intended to be used with coverlet and should not cause any program level exceptions.

Which portion of the extra credit was uncompleted? As far as I can see, The capacity of the ConsoleWriter to write to console has been tested and the JsonJokeService is able to retrieve and translate jokes from the JSON encoding of the API.

Finally, note that the way you conducted the peer review has marked you as a contributor, and not as a reviewer

GavenElder:
Hello!
What I meant by "exception" is that it looks rather complicated, your code works great with it! I probably should have clarified or used a different word rather than "exception": to describe that! 

I went back and ran your extra credit portion and you are right, it does run as intended. It must; have been something on my end!

Cheers!

A-Tanner:
That all makes sense, no worries. Thanks for getting back to me about the specifics and even re-running my code!

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -2,14 +2,22 @@
...
-        private HttpClient HttpClient { get; } = new();
+
+        public string? URL { get; private set; }
+        protected HttpClient HttpClient { get; } = new();
+
 
         public string GetJoke()
         {
-            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
+            string joke = HttpClient.GetStringAsync(URL).Result;
```
mmwoodfo, line 14: Although `GetStringAsync` allows for a null string, you may still want to do a null check here. I'd imagine you would expect the user to call `SetupService` before calling this method?

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -3,6 +3,11 @@
   <PropertyGroup>
```
mmwoodfo, line 3:
```c#
        static void Main()
        {
            //Feel free to use your own setup here - this is just provided as an example
            var jokeService = new JokeService();
            jokeService.SetupService("https://geek-jokes.sameerkumar.website/api");
            new Jester(jokeService, new ConsoleOutput()).TellJoke();
        }
```

You should configure your `Main` method in `Program.cs` to do something when the program starts up. For example, the above would be acceptable. Or you can call the json service.

CanHazFunny/CanHazFunny/JsonJokeService.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            _Joke? joke = JsonSerializer.Deserialize<_Joke?>(HttpClient.GetStringAsync(URL).Result);
+            return joke?.joke;
+        }
+
+
+        private class _Joke
+        {
+            public string? joke { get; set; }
+        }
```
mmwoodfo, line 24: I would move this class to be outside of the JsonJokeService class. Aside from that, `joke` should be pascal case. The issue of course, is it doesn't deserialize properly that way.

To fix that you could import `Newtonsoft.Json` and do the following...
```suggestion
        private class _Joke
        {
            [JsonProperty("joke")]
            public string? Joke { get; set; }
        }
```

PR #81, opened 2022-02-02
Assignment3

MarkCapka:
@TylerRose8 - partner:
Note: can't automatically merge, but won't tell us why (probably a write access thing to Intellitect's repo and we aren't supposed to merge?)

 Everything seems to work fine. 

Project works as intended.


CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,28 @@
+﻿
+namespace CanHazFunny;
+using System;
+
+
+
+public class Jester : IJesterOutput, IJokeService
```
mmwoodfo, line 7: Jester shouldn't inherit these interfaces, but rather take them as dependencies (in a constructor).

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -1,14 +1,18 @@
...
 {
-    public class JokeService
+    public class JokeService : IJokeService
     {
         private HttpClient HttpClient { get; } = new();
 
         public string GetJoke()
         {
-            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
+            string joke = "Coroners refer to dead people as \"ABC's\". Already Been Chucked."; //Favorite joke   (one that always fails ;)
```
mmwoodfo, line 9: It may be better to use a do while loop here rather than setting a 'failure' case.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+
+    private JokeService JokeService { get { return new(); } }
```
mmwoodfo, line 10: This should be getting set in a constructor.

PR #94, opened 2022-02-05
Assignment4

A-Tanner:
Extra Credit not yet attempted.

Given the loose recommendations of the assignment, it seemed like good practice in SOLID principles to create a public, abstracted class to manage Nodes and access their data.

## Notes to Grader, Peer reviewer:

**Node is a protected class of CircularlyLinkedList**
**Clear is implemented in the CircularlyLinkedList class**
**Exists is implemented in the CircularlyLinkedList class**
**Error for duplicate append is thrown by CircularlyLinkedLists.Insert()**

### DIFF
--- a/GenericsHomework/CircularlyLinkedList.Tests/CircularlyLinkedListTests.cs
+++ b/GenericsHomework/CircularlyLinkedList.Tests/CircularlyLinkedListTests.cs
@@ -1,0 +1,15 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace GenericsHomework.Tests
+{
+    [TestClass]
+    public class CircularlyLinkedList
+    {
+        [TestMethod]
+        public void TestMethod1()
+        {
+        }
+    }
+
+
+}
--- a/GenericsHomework/CircularlyLinkedList.Tests/CircularlyLinkedListTests.cs
+++ b/GenericsHomework/CircularlyLinkedList.Tests/CircularlyLinkedListTests.cs
@@ -1,15 +1,198 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
 
 namespace GenericsHomework.Tests
 {
     [TestClass]
-    public class CircularlyLinkedList
+    public class CircularlyLinkedListTests
     {
+        int[] testIntData = { 1, 7, 22, 48 };
+        string[] testStringData = { "Diego", "Karolyn", "Meatball Supreme" };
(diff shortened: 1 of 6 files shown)


GavenElder:
I will PR this.

GavenElder:
**Instructions**
        Define the Node class
            **The Next property setter should be private. ❌ (it is set to public, not private)**
            Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ✔
        Create an Exists method to test to see if a value exists in the list.✔



**Extra Credit**

**(Stated above that the extra credit hasn't been attempted yet, so I wont mark these as done.)**


**Fundamentals**

    Ensure you enable:
        net6 targetted ✔
        C# 10.0 ✔

GenericsHomework/CircularlyLinkedList.Tests/CircularlyLinkedListTests.cs
```diff
@@ -0,0 +1,190 @@
...
+        string[] testStringData = { "Diego", "Karolyn", "Meatball Supreme" };
+        string?[] testNullData = { "The", "Data", null, "Big" };
+
+
+        [TestMethod]
+        public void Exists_IfDataNotInList_ReturnsFalse()
+        {
+            CircularlyLinkedList<int> cll = new();
+            cll.AddData(testIntData);
+            Assert.IsFalse(cll.Exists(42));
```
mmwoodfo, line 20: Rather than doing the `Assert` on the method call directory, you may consider saving the result in a variable and calling the assert on the variable holding the method's result. This is mainly to help during debugging, it's nice when you can set a breakpoint on the assert statement and see what is going into it before the assert is evaluated.

GenericsHomework/CircularlyLinkedList.Tests/CircularlyLinkedListTests.cs
```diff
@@ -0,0 +1,190 @@
...
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentException))]
+        public void AddData_IfDuplicateData_ThrowsException()
+        {
+            CircularlyLinkedList<string> cll = new();
+            cll.AddData("fred");
+            cll.AddData("fred");
+        }
+        [TestMethod]
+        public void AddData_IncrementsCoundt()
```
mmwoodfo, line 39:
```suggestion
        public void AddData_IncrementsCount()
```

mmwoodfo (commented):
Great job

PR #95, opened 2022-02-06
Assignment4

sammyboy2000:


MikePHaynes:
Will review this!

MikePHaynes (commented):
It seems that while you technically covered most of the requirements you did so without the use of generics, which adds unneeded complexity to your code as well as defeats one of the purposes of this assignment. Additionally, in the "Append" method it seems that you are appending the node at the end of the list. My understanding was that we were supposed to append the node after the current node (the node that called the Append method) but please correct me if I'm wrong. No messages from the compiler which was terrific!

## Instructions
  - Define the `Node` class
    - The `Next` property setter should be private. ❌
  - Add an `Append` method that takes a value and appends a new `Node` instance after the current node (by invoking the `Next` property). ❌

## Extra Credit

1. Implement a `VennDiagram` structure that contains `n` `Circle`s that only contains homogenous **reference types** of any type. ❌

- Each circle contains n items and each item can belong to one more more `Circle` instances.
- You are not required to use a `Node` from earlier in the homework for your venn diagram implementation.
- You are welcome to use exising BCL generic classes for the extra credit.

2. Implement `Systm.Collections.Generic.ICollection<T>` on the `Node` class ❌

## Fundamentals
- Ensure you enable:
  - net6 targetted  ✔
  - C# 10.0  ✔
- For this assignment, always use `Assert.AreEqual<T>()` (the generic version)  ❌
- Choose simplicity over complexity ❌

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    public class NodeTests
+    {
+        [TestMethod]
+        public void Node_ReturnsNode()
+        {
+            Node node = new(null);
+            Assert.IsNotNull(node);
+        }
+        [TestMethod]
+        public void ToString_ProperlyParses()
```
MikePHaynes, line 15: I don't think you should have to parse anything, a generic structure should take care of that.

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,61 @@
...
+            Assert.AreEqual("39", node.ToString());
+            node.Value = (Int64)64;
+            Assert.AreEqual("64", node.ToString());
+            node.Value = 42.1;
+            Assert.AreEqual("42.1", node.ToString());
+            node.Value = new Node(null);
+            Assert.AreEqual("Type unknown", node.ToString());
+        }
+        [TestMethod]
+        public void Append_ProperlyAppends()
```
MikePHaynes, line 33: You could probably simplify this test method a little bit, I have a similar test method in my own code.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
+﻿namespace GenericsHomework
+{
+    public class Node
```
MikePHaynes, line 3: Should be something like "public class Node<T>".

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+    {
+        public object? Value { get; set; }
+        private Node Next { get;  set; }
+        public Node GetNext() { return Next; }
```
MikePHaynes, line 7: Should be private set on Next and you probably don't need the GetNext method.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+        public Node(object? value)
+        {
+            Value = value;
+            Next = this;
+        }
+        public override string ToString()
```
MikePHaynes, line 13: This is where the "complexity" part comes in. Using generics you won't have to check the type you will just need to call ToString() on Value.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+            {
+                return $"{(Int64)Value}";
+            }
+            if(type == typeof(float) || type == typeof(Double))
+            {
+                return $"{(Double)Value}";
+            }
+            return "Type unknown";
+        }
+        public void Append(object? value)
```
MikePHaynes, line 44: Should append after current Node is my understanding. Again, please correct me if I'm wrong.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
+﻿namespace GenericsHomework
+{
+    public class Node
+    {
+        public object? Value { get; set; }
```
MikePHaynes, line 5: Should be something like "Public T Value { get; set; }"

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+        private Node Next { get;  set; }
+        public Node GetNext() { return Next; }
+        public Node(object? value)
```
MikePHaynes, line 8: T Value

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+            Node index = this;
+            while (Next != index)
+            {
+                index = index.Next;
+            }
+            index.Next = new Node(value);
+            index = index.Next;
+            index.Next = this;
+        }
+        public bool Exists(object? Value)
```
MikePHaynes, line 57: T Value

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            Node node = new("1st");
+            node.Append("2nd");
+            node.Append("3rd");
+            Assert.AreEqual("1st", node.ToString());
+            Assert.AreEqual("2nd", node.GetNext().ToString());
+            Assert.AreEqual("3rd", node.GetNext().GetNext().ToString());
+            Assert.AreEqual("1st", node.GetNext().GetNext().GetNext().ToString());
```
MikePHaynes, line 41: Assert.AreEqual<string>

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+        [TestMethod]
+        public void Clear_MemoryCleared()
+        {
+            Node node = new("1st");
+            node.Append("2nd");
+            node.Append("3rd"); //Check memory usage here, using Diagnostic Tools
+            node.Clear();
+            GC.Collect();   
+            Assert.AreEqual(node, node.GetNext());//Check memory usage here, is lower
```
MikePHaynes, line 58: Assert.AreEqual< Node >

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,61 @@
...
+            node.Value = false;
+            Assert.AreEqual("false", node.ToString());
+            node.Value = 39;
+            Assert.AreEqual("39", node.ToString());
+            node.Value = (Int64)64;
+            Assert.AreEqual("64", node.ToString());
+            node.Value = 42.1;
+            Assert.AreEqual("42.1", node.ToString());
+            node.Value = new Node(null);
+            Assert.AreEqual("Type unknown", node.ToString());
```
MikePHaynes, line 30: "Assert.AreEqual< string > (42, node);" Shouldn't have to call ToString should automatically be invoked.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,58 @@
...
+            Next = this;
+        }
+        public override string? ToString()
+        {
+            return Value.ToString();
+        }
+        public void Append(T value)
+        {
+            if (Exists(value))
+                throw new ArgumentException("Value already exists in list.");
```
mmwoodfo, line 20:
```suggestion
                throw new ArgumentException($"{nameof(value)} already exists in list.");
```
Consider using `nameof` here

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,58 @@
...
+            Node<T> index = this;
+            while (Next != index)
+            {
+                index = index.Next;
+            }
+            index.Next = new Node<T>(value);
+            index = index.Next;
+            index.Next = this;
+        }
```
mmwoodfo, line 29: I don't think this works the way you expect it to...
Lets say I do the following (based off of your unit tests)
```c#
 Node<int> node = new(1);
node.Append(2);
node.Append(3);
// Your list now looks like this: 1-> 2-> 3 Great, that works. However, when you try to append in the middle of the list 
//                                ┖───────┘
things get weird. Take the following...     
node = node.Next; // node now = 2... So our location in the list is the following
// 1 -> 2 (our location) -> 3
// ┖────────────────────────┘
// Now, lets append 4. I think what you intend to do is make the following list:
// 1 -> 2 -> 3 -> 4
// ┖──────────────┘
node.Append(4);
// But what actually happens is this:
// 1 -> 2 -> 3 -> 4
//      ┖─────────┘
// 1 is now left out of the loop.

// You made this a little more complex than what was asked in the homework. The homework only asks you to insert after the current node (not at the end) so you could simply do the following:
var temp = Next;
var node = new Node<T>(value) { Next = temp };
Next = node;
// This will give you the list // 1 -> 3 -> 4 -> 2
//                                ┖──────────────┘
// Of course, the ordering is odd since we are inserting after the current node rather than at the very end. To insert at the very end you'd need to keep track of the head node, which is quite difficult to do in a circular list (as there isn't really a head node). 
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,58 @@
...
+            //Does work when simply setting this to this.Next
+            //GC is smart enough to clean up a circling list that has no references
+            //debug builds will hang onto that memory for longer though.
+            if (this == this.Next)
+               return;
+           Node<T> index = this;
+           while (index.Next != this)
+               index = index.Next;
+           index.Next = index.Next.Next;
+            this.Next = this;
```
mmwoodfo, line 55: You can simply do 
`Next = this`

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+            index.Next = new Node<T>(value);
+            index = index.Next;
+            index.Next = this;
+        }
+        public bool Exists(T Value)
+        {
+            Node<T> temp = new(Value);
+            Node<T> index = this;
+            if(temp.ToString() == index.ToString())
```
mmwoodfo, line 34: You should use `Equals` here - not `ToString`

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        [TestMethod]
+        public void Append_ProperlyAppends()
+        {
+            Node<string> node = new("1st");
+            node.Append("2nd");
+            node.Append("3rd");
+            Assert.AreEqual<string>("1st", node.ToString()!);
+            Assert.AreEqual<string>("2nd", node.Next.ToString()!);
+            Assert.AreEqual<string>("3rd", node.Next.Next.ToString()!);
+            Assert.AreEqual<string>("1st", node.Next.Next.Next.ToString()!);
```
mmwoodfo, line 35: You should test inserting when you are focused on a node that is in the middle of the linked list

PR #96, opened 2022-02-07
Assignment4

tomclark-mtcsw:


sammyboy2000:
I will review this.

sammyboy2000:
# EWU-CSCD371-2021-Winter

## See [Docs](Docs)

## Assignment 4
The purpose of this assignment is to write a class library is to learn how to write a generic class and a generic method. To accomplish this you will write a linked list that circles back on itself. We will also look at exception throwing and handling.

## Reading
Chapter 11, 12

## Instructions
  - Define the `Node` class
    - The `Next` property setter should be private. ✔
  - Add an `Append` method that takes a value and appends a new `Node` instance after the current node (by invoking the `Next` property). ✔

## Extra Credit

1. Implement a `VennDiagram` structure that contains `n` `Circle`s that only contains homogenous **reference types** of any type. ❌

- Each circle contains n items and each item can belong to one more more `Circle` instances.
- You are not required to use a `Node` from earlier in the homework for your venn diagram implementation.
- You are welcome to use exising BCL generic classes for the extra credit.

2. Implement `Systm.Collections.Generic.ICollection<T>` on the `Node` class ❌

## Fundamentals
- Ensure you enable:
  - net6 targetted  ✔
  - C# 10.0  ✔
- For this assignment, always use `Assert.AreEqual<T>()` (the generic version)  ✔

Seems like you've got it all together. I'd run code analysis once to fix some spacing in UnitTest1. Might also want to rename it to NodeTests.cs.

PR #97, opened 2022-02-07
Assignment4

Ashrom0:


GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,68 @@
...
+                Node <TValue> nn = new(value);
+                Node<TValue> cur = Next;
+
+                while(cur.Next != this)
+                {
+                    cur = cur.Next;
+                }
+
+                cur.Next = nn;
+                nn.Next = this;
```
mmwoodfo, line 37: This works, but you may be able to simplify it by doing the following: 
```suggestion
            if (Exists(value))
            {
                throw new ArgumentException("Value exist in the list.");
            }
            el
            Node<T> newNode = new(value);
            newNode.Next = Next;
            Next = newNode;
            return newNode;
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public Node(TValue value)
+        {
+            Value = value;
+            Next = this;
+        }
+
+        public void Clear()
+        {
+            this.Next = this;
```
mmwoodfo, line 17: Great! You should include a comment as to why this works though.
```suggestion
            Next = this;
```

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,49 @@
...
+            node.Append("Testing2");
+            node.Append("Testing3");
+            node.Append("Testing4");
+            node.Append("Testing5");
+            Assert.AreEqual<string>("Testing1", node.ToString()!);
+            Assert.AreEqual<string>("Testing2", node.Next.ToString()!);
+            Assert.AreEqual<string>("Testing3", node.Next.Next.ToString()!);
+            Assert.AreEqual<string>("Testing4", node.Next.Next.Next.ToString()!);
+            Assert.AreEqual<string>("Testing5", node.Next.Next.Next.Next.ToString()!);
+            Assert.AreEqual<string>("Testing1", node.Next.Next.Next.Next.Next.ToString()!);
```
mmwoodfo, line 28: You should create a test where you try inserting into the list in different locations. (e.g, the middle of the list, end of the list, start of the list).

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        [TestMethod]
+        public void ExistingValueAppend()
+        {
+            Node<string> node = new("value");
+            node.Append("value");
```
mmwoodfo, line 35: Shouldn't this be throwing an exception?

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,68 @@
...
+            Node<TValue> cur = Next;
+
+            while(cur != head)
+            {
+                if (cur.Value.Equals(value))
+                {
+                    return true;
+                }
+                cur = cur.Next;
+            }
```
mmwoodfo, line 54: This won't work if there is only one item in the list. Consider making this into a do-while loop.

PR #98, opened 2022-02-07
Assignment 4

TechGlue:
Initial PR
Pair Programming Partner: Nicholas Johnson @NicholasJohnsonDev

### DIFF
--- a/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -69,7 +69,7 @@ public class NodeTests
         Assert.IsTrue(headNode.Exists(21));
         Assert.IsTrue(headNode.Exists(22));
         Assert.IsTrue(headNode.Exists(20));
-        Assert.AreEqual(6,headNode.Size);
+        Assert.AreEqual(6, headNode.Size);
     }
 
     [TestMethod]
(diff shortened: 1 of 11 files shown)


GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,121 @@
...
+        output.Append(head.Value);
+
+        while (currentNode != head)
+        {
+            output.Append(" -> ");
+            output.Append(currentNode.Value);
+            currentNode = currentNode.Next;
+        }
+
+        return output.ToString();
```
mmwoodfo, line 119: This should technically return the default `ToString` implementation for the value stored in `Value`, not the entire linked list printed out and all of their values. However, you did a good job implementing it the way you did. :)

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,218 @@
...
+        Node<int> headNode = new(42);
+
+        //Act
+        headNode.Append(22);
+        headNode.Append(21);
+        
+        //Assert
+        Assert.AreEqual(headNode, headNode.Next.Next.Next);
+        Assert.AreEqual(22, headNode.Next.Next.Value);
+    }
```
mmwoodfo, line 126: You could probably combine these two tests

mmwoodfo (commented):
Great job

PR #99, opened 2022-02-08
Assignment4

GavenElder:
final submission.


A-Tanner (changes requested):
Missing a comment about garbage collection, otherwise looks ok. 

Some of the methods/functionality in CircularlyLinkedList appears to have been copy-and-pasted from my own codebase and that's really upsetting. 


Define the Node class
Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌ --> No explanatory comment found
Extra Credit

Fundamentals
Ensure you enable:
net6 targetted ✔
C# 10.0 ✔

Assignment4/GenericsHomework/CircularlyLinkedList.Tests/CLLTests.cs
```diff
@@ -0,0 +1,178 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
+namespace GenericsHomework.Tests
+{
+    [TestClass]
+    public class CLLTests
+    {
+        int[] IntData = { 1, 2, 3, 4 };
```
mmwoodfo, line 9: These variable names don't match the variable names you're using in your tests. Causing your code to not compile.

Assignment4/GenericsHomework/GenericsHomework/CircularlyLinkedList.cs
```diff
@@ -0,0 +1,142 @@
+using System.Diagnostics.CodeAnalysis;
+
+namespace GenericsHomework;
+public class CircularlyLinkedList<T>
+{
+    protected Node<T>? _Cursor { get; set; }
```
mmwoodfo, line 6: For the code review, take a look at PR #94 as the code is similar enough that the review is the same...

mmwoodfo (commented):
Your code does not compile and therefore cannot be properly graded. Additionally, your code looks similar to another student's codebase.

PR #100, opened 2022-02-08
Assignment4

NazariyPogo:
Made node class.
Made tests for node class.
Unit tested node class.


czamora7:
I will review this

GenericsHomework/GenericsHomeworkTests/GenericsHomeworkTests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
czamora7, line 3: .NET analyzers should be enabled here

GenericsHomework/GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace GenericsHomeworkTests
+{
+    [TestClass]
+    public class NodeTests
+    {
+        [TestMethod]
+        public void Node_ConstructorWorks()
+        {
+            Node<string> node = GetNode();
+            Assert.IsNotNull(node);
```
czamora7, line 14: Assert.AreEqual<T> should be used for this assignment

czamora7 (commented):
Define the Node class
Whether it is sufficient to only set Next to itself ❌
Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌
Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌

Fundamentals
Ensure you enable:
net6 targetted ✔
C# 10.0 ✔
and enabled .NET analyzers for both projects ❌

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public override string? ToString()
+        {
+            if(this.Value is null)
+            {
+                return null;
+            }
+
+            return this.Value.ToString();
```
mmwoodfo, line 66:
```suggestion
return Value.ToString();
```
The above change will do the same thing.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,69 @@
...
+            if (this.Value.Equals(value)) { return true; }
+
+            Node<T> tracer = this;
+
+            for (; tracer.Next != this; tracer = tracer.Next)
+            {
+                if(tracer.Value.Equals(value)) { return true; }
+            }
+
+            if (tracer.Value.Equals(value)) { return true; }
```
mmwoodfo, line 54: You can also do this via a do-while loop, which may be cleaner to read.

GenericsHomework/GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            Node<string> node = GetNode();
+            node.Append("Princess Buttercup");
+            Assert.AreEqual<string>("Inigo Montoya", node.Value);
+        }
+
+        [TestMethod]
+        public void Node_Append_NewValueAppended()
+        {
+            Node<string> node = GetNode();
+            node.Append("Princess Buttercup");
```
mmwoodfo, line 43: You should write some tests for inserting multiple values, inserting into the 'middle' (when i say middle, I mean change your node to not be the first one created) of the list, etc.

mmwoodfo:
This coebase looks similar to #101

PR #101, opened 2022-02-08
Assignment4

xiokd:
Code Pair: @dhoxie

### DIFF
--- a/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -1,0 +1,14 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace GenericsHomework.Tests
+{
+    [TestClass]
+    public class NodeTests
+    {
+        [TestMethod]
+        public void TestMethod1()
+        {
+
+        }
+    }
+}
--- a/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -5,6 +5,9 @@ namespace GenericsHomework.Tests
     [TestClass]
     public class NodeTests
     {
+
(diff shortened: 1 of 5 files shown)


dhoxie:
## Review
  - Define the `Node` class
    - The `Next` property setter should be private. ✔
  - Add an `Append` method that takes a value and appends a new `Node` instance after the current node (by invoking the `Next` property). ✔

## Extra Credit
We have not attempted extra credit

## Fundamentals
- Ensure you enable:
  - net6 targetted  ✔
  - C# 10.0  ✔
- For this assignment, always use `Assert.AreEqual<T>()` (the generic version)  ✔

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
...
+            while(currentNode.Next != this) { currentNode = currentNode.Next; }
+
+            currentNode.Next = newNode;
+            newNode.Next = this;
+        }
+
+        public void Clear()
+        {
+            // don't need to garbage collect as c# does so automatically
+            Next = this;
```
mmwoodfo, line 36: But how does C# know what to garbage collect automatically? Your implementation is correct, but I think you need to elaborate on it.

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            Assert.AreEqual<string?>(null, result);
+        }
+
+        [TestMethod]
+        public void Append_GivenNewValues_Success()
+        {
+            Node<string> node = GetNodes();
+
+            Assert.IsTrue(node.Exists("Baggy"));
```
mmwoodfo, line 34: Try writing a test that inserts multiple values

dhoxie: The GetNodes() method inserts multiple values

PR #102, opened 2022-02-08
Assignment4

whitneybolar:
When I pushed the work over to Github, my work did not fully update. I made a zip file with everything inside. Please feel free to run it. I added each file I edited for easy access.


Node.cs
```diff
@@ -0,0 +1,96 @@
+﻿namespace GenericsHomework
+{
+    public class Node
+    {
+        public object? Value { get; set; }
+
+        public Node Next { get; private set; }
+        public Node? Head { get; set; }
```
mmwoodfo, line 8: You shouldn't need to be keeping track of the head node.

Node.cs
```diff
@@ -0,0 +1,96 @@
+﻿namespace GenericsHomework
+{
+    public class Node
```
mmwoodfo, line 3: This should be generic, using `<T>`

Node.cs
```diff
@@ -0,0 +1,96 @@
...
+        public object? Value { get; set; }
+
+        public Node Next { get; private set; }
+        public Node? Head { get; set; }
+
+        public Node? Tail{ get; set; }
+
+        public int Size;
+
+        public Node(object? value)
```
mmwoodfo, line 14:
```suggestion
        public Node(T value)

```

Node.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        public Boolean Exists(object? value)
+        {
+            Node cur = new Node(value);
+            Node i = this; // an index
+
+               if (this.Next == this) // wraps around and stops at the current node 
+                {
+                   if (cur.ToString() == i.Next.ToString()) {
```
mmwoodfo, line 60: This should be using `.Equals`, not `ToString`

Node.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                Head = cur;
+                Tail = cur;
+            }
+            else
+            {
+                Head = cur.Next;
+            }
+
+            Size++;
```
mmwoodfo, line 49: This doesn't append anything to your list, your next value never changes

UnitTest1.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+
+        [TestMethod]
+        public void Node_Append()
+        {
+            Node cur = new Node("First Node");
+            cur.Append("obj");
+            Assert.AreEqual(cur.Size, cur.GetSize());
+            Assert.AreEqual("First Node", cur.Next.ToString());
```
mmwoodfo, line 25: If you've appended to the list, why would `cur.Next` still point to it's self?

Node.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+           
+        
+        public void Clear()
+        {
+            Node i = this; 
+
+            while (i.Size != 1) // there should only be one left everytime 
+            {
+                this.Next = this;
```
mmwoodfo, line 75: Why put `this.Next = this` inside a loop when `this.Next`'s value doesn't change? Would you not want to delete the while loop and simply set size to 0? As this loop does not do anything usful.

Node.cs
```diff
@@ -0,0 +1,96 @@
...
+            Size++;
+
+        }
+
+        public Boolean Exists(object? value)
+        {
+            Node cur = new Node(value);
+            Node i = this; // an index
+
+               if (this.Next == this) // wraps around and stops at the current node
```
mmwoodfo, line 58: Don't you want to check every node to make sure a duplicate doesn't exist anywhere in the list?

mmwoodfo:
You didn't use generics anywhere in this assignment

PR #103, opened 2022-02-08
Assignment4

ericlplanteschool:
implemented base requirements, hit a snag with unit tests

### DIFF
--- a/GenericsHomework/GenericsHomework/GenericsHomework.cs
+++ b/GenericsHomework/GenericsHomework/GenericsHomework.cs
@@ -28,11 +28,22 @@
 			if(Exists(args)){
 				throw new ArgumentOutOfRangeException();
 			}
+
+			Node<T> head = next;
 			
 			next = new Node<T>(args);
+
+			next.next = head;
+
 		}
 
+		//Because this is not a circular list and it does not contain and references to the previous node, 
+		//when reassigned as the previous members of the list will fall out of scope and therefor we do not need to unassign each of them.
 		public void Clear(){
+			if(next == this){
+				return;
+			}
+
 			Node<T> cur = next;
 
 			while(cur.next != this){
@@ -57,7 +68,9 @@
 			do
 			{
 				if(sameType(cur.data, args)){
-					return true;
+					if(cur.data.Equals(args)){
+						return true;
+					}
 				}
 
 			}while(cur.next != this);
--- a/GenericsHomework/GenericsHomework/GenericsHomework.cs
+++ b/GenericsHomework/GenericsHomework/GenericsHomework.cs
@@ -1,0 +1,68 @@
+﻿namespace GenericsHomework
+{
+	public class Node<T>
+		where T : notnull
+	{
+		private T data;
+		private Node<T> next;
+
+		public Node(T parms)
+		{
+			data = parms;
+			next = this;
+		}
+
+		private Node<T> Next
+        {
+            get { return next; }
+            set { next = value; }
+        }
+
+		public override string ToString()
+		{
+			return "" + data.ToString();
+		}
+
+		public void Append(T args)
+		{
+			if(Exists(args)){
+				throw new ArgumentOutOfRangeException();
+			}
+			
+			next = new Node<T>(args);
+		}
+
+		public void Clear(){
+			Node<T> cur = next;
+
+			while(cur.next != this){
+				cur = cur.next;
+			}
+
+			Node<T> placeholder = cur.next;
+
+			cur.next = cur;
+
+			placeholder.next = placeholder;
+		}
+
+		public bool Exists(T args)
+		{
+			bool sameType(T arg1, T arg2){
+				return arg1.GetType == arg2.GetType;
+			}
+
+			Node<T> cur = this;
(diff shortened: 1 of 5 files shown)


Ashrom0:
I will review

Ashrom0 (commented):
Define the Node class
Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌

Ensure you enable:
net6 targetted ✔
C# 10.0 ✔
All of the above should be unit tested ❌

Seems that the unit test are still being worked along with the Assert stuff but overall, seems fine

GenericsHomework/GenericsHomework/GenericsHomework.cs
```diff
@@ -0,0 +1,81 @@
+﻿namespace GenericsHomework
```
mmwoodfo, line 1:
```suggestion
using System;

﻿namespace GenericsHomework
```

Your code doesn't compile because you are missing a using statement. This is needed because of the exception you are throwing further down in the code.

GenericsHomework/GenericsHomework/GenericsHomework.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+	public class Node<T>
+		where T : notnull
+	{
+		private T data;
+		private Node<T> next;
```
mmwoodfo, line 7:
```suggestion
		private T _Data;
		private Node<T> _Next;
```

GenericsHomework/GenericsHomework/GenericsHomework.cs
```diff
@@ -0,0 +1,81 @@
...
+		private Node<T> Next
+        {
+            get { return next; }
+            set { next = value; }
+        }
+
+		public override string ToString()
+		{
+			return "" + data.ToString();
+		}
```
mmwoodfo, line 24:
```suggestion
		public override string? ToString()
		{
			return data?.ToString();
		}
```

GenericsHomework/GenericsHomework/GenericsHomework.cs
```diff
@@ -0,0 +1,81 @@
...
+
+		public void Append(T args)
+		{
+			if(Exists(args)){
+				throw new ArgumentOutOfRangeException();
```
mmwoodfo, line 29: You shouldn't be throwing an out of range exception, as that's not the problem here. Simply throw an ArgumentException with a message saying that the value already exists, or something along those lines.

GenericsHomework/GenericsHomework/GenericsHomework.cs
```diff
@@ -0,0 +1,81 @@
...
+
+			Node<T> head = next;
+			
+			next = new Node<T>(args);
+
+			next.next = head;
+
+		}
+
+		//Because this is not a circular list and it does not contain and references to the previous node,
```
mmwoodfo, line 40: This is a circular list.

GenericsHomework/GenericsHomeworkTests/GenericsHomeworkTests.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+	[TestClass]
+	public class GenericsHomework
+	{
+
+		[TestMethod]
+		public void Constructor()
+		{
+			// Arrange.
+			//Node<T> node = new Node("") missing using directive?
```
mmwoodfo, line 15: You're having trouble because you never added a project reference to your `GenericsHomework` project.

GenericsHomework/GenericsHomework/GenericsHomework.cs
```diff
@@ -0,0 +1,68 @@
...
+			}
+
+			Node<T> placeholder = cur.next;
+
+			cur.next = cur;
+
+			placeholder.next = placeholder;
+		}
+
+		public bool Exists(T args)
```
mmwoodfo, line 60: args isn't the best name for a variable - as it suggests multiple values are being passed in.

GenericsHomework/GenericsHomework/GenericsHomework.cs
```diff
@@ -0,0 +1,68 @@
...
+		{
+			bool sameType(T arg1, T arg2){
+				return arg1.GetType == arg2.GetType;
+			}
```
mmwoodfo, line 64: You don't need this here. Simply use `Equals`

GenericsHomework/GenericsHomework/GenericsHomework.cs
```diff
@@ -0,0 +1,68 @@
...
+
+			Node<T> cur = this;
+			
+			do
+			{
+				if(sameType(cur.data, args)){
+					return true;
+				}
+
+			}while(cur.next != this);
```
mmwoodfo, line 76: This is an endless loop, you need to be setting cur equal to next so you can move through the list items.

mmwoodfo:
I don't see a PR review, did you do one?

ericlplanteschool:
> I don't see a PR review, did you do one?

no i forgot to complete a peer review

PR #104, opened 2022-02-08
Assignment4

czamora7:


NazariyPogo:
I'll look into this.

NazariyPogo:
Define the Node class
Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌ //no comment provided

Extra Credit

Fundamentals
Ensure you enable:
net6 targetted ✔
C# 10.0 ✔

GenericsHomework/GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,56 @@
...
+    public void Init()
+    {
+        Node<object> nn = new Node<object>(1);
+        _Node = nn;
+    }
+
+    [TestMethod]
+    public void TestToStringReturnsCorrectly()
+    {
+        Assert.AreEqual<object>(_Node!.ToString(), "1");
```
mmwoodfo, line 22:
```suggestion
        Assert.AreEqual<string>(_Node!.ToString(), "1");
```
Using `object` sort of defeats the point of generics.

GenericsHomework/GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,56 @@
...
+    [TestMethod]
+    public void TestNextInitialisesCorrectly()
+    {
+        Assert.AreEqual<GenericsHomework.Node<object>>(_Node!.Next, _Node);
+    }
+
+    [TestMethod]
+    public void TestAppendMethodAddsValueCorrectly()
+    {
+        _Node!.Append(2, _Node);
```
mmwoodfo, line 34: You should try appending multiple values and changing positions in the list while testing the append method.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,81 @@
...
+            head.Next = head;
+        }
+    }
+
+    public bool Exists(T data, Node<T> root)
+    {
+        Node<T> check = root;
+        while (check.Next != check && check.Next != root)
+        {
+            if (new Node<T>(data).CompareTo(check) == 0)
```
mmwoodfo, line 68: You should be able to just use `Equals` here. When checking for duplicates, you just need to check the value stored in the node object against the value you pass into this method. You can use that values base `Equals` implementation.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public int CompareTo(Node<T>? other)
```
mmwoodfo, line 77: You shouldn't be using `CompareTo` here, as this is typically used for sorting, not usually checking equality.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,81 @@
...
+            head.Next = head;
+        }
+        //else the list is circular
+        else
+        {
+            //set the tails next to self reference
+            check.Next = check;
+            //set the heads next to self reference
+            head.Next = head;
+        }
```
mmwoodfo, line 60: The C# garbage collector will automatically clean up anything that is no longer reachable, so anything that doesn't have an active reference to it. To clean up the list, all you should have to do is `Next = this`. You could also loop through the entire list and set all of the nodes equal to themselves, but the above line is sufficient for this assignment.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,81 @@
...
+        Data = t;
+        Next = this;
+    }
+
+    public override String ToString()
+    {
+        return Data?.ToString();
+    }
+
+    public void Append(T data, Node<T> root)
```
mmwoodfo, line 24: What is a root node in a circularly linked list? You don't need to keep track of the head here, just insert the data after the current node.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        Node<T> check = root;
+        while(check.Next != check && check.Next != root)
```
mmwoodfo, line 27: This will always throw when adding another node to a list with only one node in it. You should be using the `Exists` method here, as that should be checking for duplicates in your list and returning true or false.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            throw new Exception("duplicate value");
+        }
+
+        check.Next = new Node<T>(data);
```
mmwoodfo, line 32: This will only ever make a list 2 long, overwriting the previous `.Next` value.

PR #105, opened 2022-02-08
Assignment4

MarkCapka:
@TylerRose8 - partner

Issue with being able to create objects for tests that we need to ask about on Tuesday. Will finish out tests then as currently, untestable for me.  --Mark Capka


GenericsHomework/GenericsHomework/CircularLinkedList.cs
```diff
@@ -0,0 +1,197 @@
...
+
+        Add(value);
+        return this;
+    }
+    public bool Exists(T value)
+    {
+        return Contains(value);
+    }
+
+   // consider: Whether it is sufficient to only set Next to itself ✔      SEE LINE BELOW FOR RESPONSE:
```
mmwoodfo, line 78: 👍🏻

GenericsHomework/GenericsHomework.Tests2/CircularLinkedListTests.cs
```diff
@@ -0,0 +1,260 @@
...
+    {
+        string value = "start";
+        string value2 = "next";
+        string? value3 = null;
+
+        CircularLinkedList<string> list = new(value);
+
+        list.Append(value2).Next();
+
+        Assert.IsTrue(list.Contains(value3!));
```
mmwoodfo, line 111: I don't think `value3` is ever added - so I think you may want to be asserting that this is false.

GenericsHomework/GenericsHomework/CircularLinkedList.cs
```diff
@@ -0,0 +1,197 @@
...
+
+    public bool Remove(T item)
+    {
+        if (item is null)
+        {
+            return false;
+        }
+        Node<T> current = Cursor;
+        int searched = 0;
+        while (!current.Next.Equals(current) && searched++ < Count)
```
mmwoodfo, line 171: It doesn't appear that you are checking against the given `item` value passed into this method. Hence why your test for this fails.

GenericsHomework/GenericsHomework.Tests2/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
mmwoodfo, line 1: This project is a console project, not a class library

PR #107, opened 2022-02-08
Assignment4 with @ZackWard-student

jmplett:
@ZackWard-student

### DIFF
--- a/GenericsHomework/GenericsHomework/Node.cs
+++ b/GenericsHomework/GenericsHomework/Node.cs
@@ -2,22 +2,82 @@
 {
     public class Node<T>
     {
-        private int[] size = { 0 };
-        private Node<T> next = null!;
-        private T? Value;
-        public int Size { get => size[0]; }
+        private int[] _Size = { 0 }; //should be converted to a shared int using pointers
+        private Node<T> _Next;
+        private T _Value;
+        public T Value {
+            get
+            {
+                return _Value;
+            }
+            set
+            {
+                if (value == null)
+                {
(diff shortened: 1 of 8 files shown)


GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,97 @@
+﻿namespace GenericsHomework
+{
+    public class Node<T> where T : notnull //maybe make this nullable...
```
mmwoodfo, line 3:
```suggestion
    public class Node<T> 
```
You shouldn't really need to add this on at all.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,97 @@
...
+        public override string ToString()
+        {
+            Node<T> cursor = this;
+            string result = $"{this.Value}";
+            for (int i = 1; i < Count; i++)
+            {
+                cursor = cursor.Next;
+                result = $"{result}, {cursor.Value}";
+            }
+            return $"[{result}]";
```
mmwoodfo, line 78: This was intended to simply return the string value of the value stored in the current node... So `Value.ToString()`, but what you did works as you intended it too, so it's not wrong in the context of your code.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        public void Clear()
```
mmwoodfo, line 81: There should be a comment here explaining why this works.

PR #108, opened 2022-02-09
Assignment4 @ Tony Chan

tnadalini:


GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,84 @@
...
+    public Node(T data)
+    {
+        Data = data;
+        Next = this!;
+    }
+
+    public override string ToString()
+    {
+        return Data?.ToString()!;
+    }
```
mmwoodfo, line 24:
```suggestion
    public override string? ToString()
    {
        return Data?.ToString();
    }
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,84 @@
+﻿namespace GenericsHomework;
+
+using System;
+public class Node<T>
+{
+    private Node<T>? _Next;
```
mmwoodfo, line 6: This should be marked as non-nullable

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,84 @@
...
+    {   
+        get { return _Next!; }
+        private set { _Next = value!; }
+    }
+
+    public Node(T data)
+    {
+        Data = data;
+        Next = this!;
+    }
```
mmwoodfo, line 19:
```suggestion
    public Node(T data)
    {
        Data = data;
        _Next = this;
    }
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            current = current.Next;
+
+        } while (current != this);
+
+        return false;
+    }
+
+    public void Clear()
```
mmwoodfo, line 61: You should provide a comment here explaining this.

PR #109, opened 2022-02-10
Assignment4 @Disciple4GT

nwnel-017:


GenericsHomework/GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
mmwoodfo, line 4: Make sure you are using language 10.0, .NET analyzers and treat warnings as errors

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            Node<TValue> newNode = new(item);
+            Node<TValue> currentNode = Next;
+
+            while(currentNode != this)
+            {
+                currentNode = currentNode.Next;
+            }
+            currentNode.Next = newNode;
+            newNode.Next = this;
```
mmwoodfo, line 29: This will only append one additional item, as you are overwriting this's `Next` value.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        public void Clear()
```
mmwoodfo, line 32: You should provide a comment explaining why this works.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+            
+            while(currentNode.Equals(this) == false)
+            {
+                if(currentNode.Item != null && currentNode.Item.Equals(item))
+                {
+                    return true;
+                }
+                currentNode = currentNode.Next;
+            }
+            return false;
```
mmwoodfo, line 53: You can also do this with a do-while loop

PR #110, opened 2022-02-11
Assignment4 David Cox @hgill17

dcox2005:
Assignment4

### DIFF
--- a/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -103,17 +103,18 @@ public class NodeTests
     public void Clear_CheckThatGarbageCollectionRemovedOld_Success()
     {
         Node<string> myNode = createNodeList();
-        Process currentProcess = Process.GetCurrentProcess();
-        long fullListMemory = currentProcess.PrivateMemorySize64;
-
+        long fullListMemory = GC.GetTotalMemory(false);
+        
         myNode = myNode.Clear();
 
         GC.Collect();
         GC.WaitForPendingFinalizers();
 
-        long clearedListMemory = currentProcess.PrivateMemorySize64;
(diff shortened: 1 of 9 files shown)


GenericsHomework/GenericsHomework/Program.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GenericsHomework;
+
+class Program
+{
+    static void Main(string[] args)
+    {
+        Console.WriteLine(args[1]);
```
mmwoodfo, line 13: This is supposed to be a class library, not a console application.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,83 @@
...
+                return true;
+            }
+            else
+            {
+                currentNode = currentNode.Next;
+            }
+
+        } while (currentNode != this);
+        
+        return false;
```
mmwoodfo, line 81: A slightly less verbose way to go about it

```suggestion
            var node = this;
            do
            {
                if (EqualityComparer<T>.Default.Equals(node.Value, value)) return true;
                node = node.Next;
            } while (node != this);
            return false;
```

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,182 @@
...
+        Node<char> myNode6 = new('H');
+        Node<float> myNode7 = new(3.14f);
+
+        Assert.IsNotNull(myNode);
+        Assert.IsNotNull(myNode2);
+        Assert.IsNotNull(myNode3);
+        Assert.IsNotNull(myNode4);
+        Assert.IsNotNull(myNode5);
+        Assert.IsNotNull(myNode6);
+        Assert.IsNotNull(myNode7);
```
mmwoodfo, line 29: You may want to check the value as well

PR #111, opened 2022-02-11
Assignment4

mpennock5:
@Aedyn01

Submission of Assignment4

### DIFF
--- a/GenericsHomework/Node.cs
+++ b/GenericsHomework/Node.cs
@@ -4,7 +4,7 @@ public class Node<TVal> where TVal : notnull
 
     public Node(TVal val)
     {
-        Val = val ?? throw new ArgumentNullException("TVal cannot be null");
+        Val = val ?? throw new ArgumentNullException($"Node<{typeof(TVal)}> value cannot be null");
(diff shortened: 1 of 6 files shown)


GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+        if(Next == this)
+        {
+            Next = node1;
+            node1.Next = this;
+        }
+        else
+        {
+            node1.Next = Next;
+            Next = node1;
+        }
```
mmwoodfo, line 31:
```suggestion
            Node<T> newNode = new(node);
            newNode.Next = Next;
            Next = newNode;
            return newNode; 
```
Another way to do this

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+            node1.Next = this;
+        }
+        else
+        {
+            node1.Next = Next;
+            Next = node1;
+        }        
+    }
+
+        /*
```
mmwoodfo, line 34: 👍🏻

PR #112, opened 2022-02-11
Assignment4 @ Dakota Condos

htran9:
final version


tremorris1999:
I will review (although too late)

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        Value = value;
+        Next = this;
+        Root = this;
+    }
+    public Node(TValue value, Node<TValue> node)
+    {
+        Value = value;
+        Next = node;
+        Root = this;
```
tremorris1999, line 18: Every node will appear to be the root!

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,75 @@
...
+        bool keyExists = Exists(value);
+        if (keyExists == true)
+        {
+            throw new ArgumentException("Cannot add duplicate value");
+        }
+        Node<TValue> lastNode = GetLast();
+        lastNode.Next = new Node<TValue>(value, Root);
+    }
+
+    public Node<TValue> GetLast()
```
tremorris1999, line 31: Clever! This could've been useful for me.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,75 @@
...
+    public void Clear()
+    {
+        // The garbage collecter picks up objects once they are no longer referenced,
+        // so removing the 'Next' reference to a node would allow the garbage collector to
+        // delete it, and subsequently delete the node it WAS pointing to because that
+        // node no longer has a reference to it and etc...
+        // 
+        // Removing a Node's Next reference effectively removes all items from that Node's list except
+        // the current node itself, because the list is now only pointing to a single node circularly (itself)
+        Root.Next = this;
```
tremorris1999, line 49: Works in this case, but if root were set properly in 2nd constructor, it would leave _at least_ 2 nodes.

tremorris1999 (changes requested):
Overall it looks good! Aside from the two issues I pointed out, there's a few things I'd question, but I think I can chalk it up to a very different implementation of the goal here.

Your tests are thorough and the implementation looks good!

Might also be worth noting you included some binary files that weren't necessary!

GenericsHomework/GenericsHomeworkTests/GenericsHomeworkTests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <Nullable>enable</Nullable>
```
mmwoodfo, line 5: Make sure you enable analyzers and treat warnings as errors for both projects

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,75 @@
+﻿namespace GenericsHomework;
+public class Node<TValue> where TValue : notnull
+{
+    public TValue Value { get; }
+    public Node<TValue> Next { get; private set; }
+    public Node<TValue> Root;
```
mmwoodfo, line 6: You shouldn't really need a `root` in a typical circular linked list.

PR #114, opened 2022-02-11
Assignment4 - Trevor Morris

tremorris1999:


GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,166 @@
...
+         */
+        private bool Exists(T value)
+        {
+            return FindIndex(value) != -1;
+        }
+
+        /**
+         * FindIndex is implicitly tested by Exists and Remove.
+         */
+        private int FindIndex(T value)
```
mmwoodfo, line 80: You can also use a do-while loop to implement this

mmwoodfo (commented):
Everything looks good, including the extra credit - however your PR review was late, so you may notice some missing points for that.

PR #116, opened 2022-02-11
Assignment3

nicoleglidden:
Shoutout to Jenny, Meg, Kevin, and Bryce for making this assignment completion possible!


CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,37 @@
...
+            JokeService = inputJokeService;
+            JokeOutput = inputJokeOutput;
+        }
+
+        public string TellJoke()
+        {
+            bool chuckNorrisJoke = false;
+            string joke = "";
+            
+            while (chuckNorrisJoke)
```
mmwoodfo, line 25: This will only enter if `chuckNorrisIsJoke` is `true`, but you start out with it being false, so you never actually call the `GetJoke` method leaving your `joke` string always empty.

nicoleglidden: Ah, shoot I see that now. I never completed the while loop. Thanks!

PR #123, opened 2022-02-16
Update dotnet.yml

BenjaminMichaelis:


BenjaminMichaelis:
Is failing because there is no assignment folder, but works as shown in #124

PR #125, opened 2022-02-16
Assignment5+6 @Tony Chan

tnadalini:


sammyboy2000:
I will review this.

sammyboy2000:

   - Include a test that leverages a hardcoded list of Spokane-based addresses. ❌
   - Include a test that uses LINQ to verify the data is sorted correctly (do not use a hardcoded list). ❌

   - Consider using `ISampleData.CsvRows` in your test to verify your results. ❌

   - Use the `people` parameter from `ISampleData.People` property for your data source. ✔
   - It is recommended that, at a minimum, you use `ISampleData.GetUniqueSortedListOfStatesGivenCsvRows` to validate your result. ❌

- Add an `IEnumberable<T> ChildItems(int maximum)` method to `Node` that returns the remaining items with a maximum number of items returned less than `maximum`.  ✔

## Extra Credit

- Implement the homework using async/await and multi-threading by defining a new `SampleDataAsync` class that implements `IAsyncSampleData`). Refactor your `SampleData` and `SampleDataAsync` classes with minimal duplication. Be sure to refactor your tests to re-use a significant amount of the test code for both implementations. ❌

## Fundamentals

- Ensure you enable:
  - net6 targeted ✔
  - C# 10.0 ✔
- All of the above should be unit tested ❌

While the code seems good, more unit tests are needed to meet the requirements. 
Unit tests are needed for the new Node methods.

nwnel-017:
I will be reviewing this one as well, since all other pull requests have been reviewed already. Looks like you need to implement the hard coded unit test using spokane addresses to test part 2. You also need to verify that part 2 is sorted correctly. Besides these issues, your project looks very good to me

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,101 @@
...
+        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+        {
+            IEnumerable<string> stateList = 
+                (
+                    from line in CsvRows
+                    let state = line.Split(',')
+                    orderby state[6]
+                    select state[6]
+                )
+                .Distinct().ToList();
```
mmwoodfo, line 42:
```suggestion
          return CsvRows
                    .Select(item => item.Split(',')[6])
                    .Distinct()
                    .OrderBy(item => item);
```
It's a bit unexpected to use the query flavor of the LINQ syntax - but if you do I would not recommend mixing the two flavors.

Above is the way to go about it with the method syntax, which is a bit more C# looking and, in my opinion, fits better in your code in terms of readability.

Additionally, you do not need to convert this to a list, as the method returns an `IEnumerable`

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,101 @@
...
+
+            return stateList;
+        }
 
         // 3.
         public string GetAggregateSortedListOfStatesUsingCsvRows()
-            => throw new NotImplementedException();
+        {
+            string[] states = GetUniqueSortedListOfStatesGivenCsvRows().Select(item => item).ToArray();
```
mmwoodfo, line 50:
```suggestion
            string[] states = GetUniqueSortedListOfStatesGivenCsvRows().ToArray();
```
You don't need this select here.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,101 @@
...
+        public IEnumerable<IPerson> People
+        {
+            get
+            {
+                IEnumerable<IPerson> people =
+                    from line in CsvRows
+                    let peopleInfo = line.Split(",")
+                    orderby peopleInfo[5], 
+                            peopleInfo[6], 
+                            peopleInfo[7]
```
mmwoodfo, line 65: This should be ordered by state, city, zip. I believe this is being ordered as city, state, zip.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,101 @@
...
+            return filteredList;
+        }
 
         // 6.
-        public string GetAggregateListOfStatesGivenPeopleCollection(
-            IEnumerable<IPerson> people) => throw new NotImplementedException();
+        public string GetAggregateListOfStatesGivenPeopleCollection(IEnumerable<IPerson> people)
+        {
+            IEnumerable<string> states = people.Select(item => item.Address.State).Distinct();
+            return states.Aggregate((state1, state2) => state1 + "," + state2);
```
mmwoodfo, line 98: You could combine this into one statement.

PR #126, opened 2022-02-20
Assignment5+6 @ Dakota Condos

htran9:
finalized SampleData class and unit tested.


Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,89 @@
...
         public string GetAggregateSortedListOfStatesUsingCsvRows()
-            => throw new NotImplementedException();
+        {
+            string[] state = GetUniqueSortedListOfStatesGivenCsvRows().Select(x => x).ToArray();
+            return string.Join(",", state);
+        }
 
         // 4.
-        public IEnumerable<IPerson> People => throw new NotImplementedException();
+        // -----Not yet tested-----. LETS HAVE A CLOSER LOOK AT THIS
```
mpennock5, line 62: This appears to work, just needs proper testing comparing sample.People to what you expect.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,89 @@
...
+        public static void Main(string[] args)
+        {
+            SampleData sampleData = new SampleData();
+            ISampleData data = new SampleData();
+            IEnumerable<string> temp = sampleData.GetUniqueSortedListOfStatesGivenCsvRows();
+            string s = sampleData.GetAggregateSortedListOfStatesUsingCsvRows();
+            IEnumerable<IPerson> people = sampleData.People;
+            IEnumerable<string> csv = sampleData.CsvRows;
+            Predicate<string> filter = email;
+            static bool email(string email) => email.Contains("jdaneltim@jimdo.com");
```
mpennock5, line 15: I would double-check that this isn't performing an unnecessary query here.

mpennock5 (commented):
Overall good job!

Assignment/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,110 @@
...
+        List <Node<TValue>> children = new();
+        Node<TValue> currentNode = Root;
+
+        int counter = 1;
+        do
+        {
+            children.Add(currentNode);
+            counter++;
+            currentNode = currentNode.Next;
+        } while (currentNode != Root && counter < maximum);
```
mpennock5, line 87: Since you are breaking when you get back to Root I believe you could also use while(moveNext() && counter < max) because it keeps track of when you have iterated over the list.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,48 @@
...
+        return CsvRows.Select(line => line.Split(',')).Select(x => x[6]).OrderBy(x => x).Distinct().ToList();
+    }
 
-        // 3.
-        public string GetAggregateSortedListOfStatesUsingCsvRows()
-            => throw new NotImplementedException();
+    // 3.
+    public string GetAggregateSortedListOfStatesUsingCsvRows()
+    {
+        string[] state = GetUniqueSortedListOfStatesGivenCsvRows().Select(x => x).ToArray();
```
mmwoodfo, line 19:
```suggestion
        string[] state = GetUniqueSortedListOfStatesGivenCsvRows().ToArray();

```
You don't need this here

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,48 @@
...
+        () => File.ReadLines(@"People.csv").Where(line => !string.IsNullOrWhiteSpace(line)).Skip(1));
+    public IEnumerable<string> CsvRows { get { return _LazyCsvRows.Value; } }
 
-        // 2.
-        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows() 
-            => throw new NotImplementedException();
+    // 2.
+    public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+    {
+        return CsvRows.Select(line => line.Split(',')).Select(x => x[6]).OrderBy(x => x).Distinct().ToList();
```
mmwoodfo, line 13:
```suggestion
        return CsvRows.Select(line => line.Split(',')[6]).OrderBy(x => x).Distinct().ToList();

```

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,48 @@
...
+            new Address(person[4], person[5], person[6], person[7]),
+            person[3]));
 
-        // 5.
-        public IEnumerable<(string FirstName, string LastName)> FilterByEmailAddress(
-            Predicate<string> filter) => throw new NotImplementedException();
+    // 5.
+    public IEnumerable<(string FirstName, string LastName)> FilterByEmailAddress(Predicate<string> filter)
+    {
+        IEnumerable<IPerson> people = new SampleData().People;
```
mmwoodfo, line 36: You shouldn't declare a new instance of sample data to get a People list here. Instead, just call the method `People`

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,48 @@
...
+        IEnumerable<(string FirstName, string LastName)> result = people.Where(x => filter(x.EmailAddress)).Select(name => (first: name.FirstName.Trim(), last: name.LastName.Trim()));
```
mmwoodfo, line 37:
```suggestion
        IEnumerable<(string FirstName, string LastName)> result = People.Where(x => filter(x.EmailAddress)).Select(name => (first: name.FirstName.Trim(), last: name.LastName.Trim()));

```

PR #130, opened 2022-02-22
Assignment5+6

schuywalker:
@Schuywalker Schuyler Asplin and @A-Tanner Ray Tanner. Monday Submission.

### DIFF
--- a/Assignment/Assignment/SampleData.cs
+++ b/Assignment/Assignment/SampleData.cs
@@ -12,22 +12,47 @@ namespace Assignment
         {
             get
             {
-                string[] temp = File.ReadAllLines("People.csv");//System.Reflection.Assembly.GetExecutingAssembly().Location; current will work but it relies on unsafe assumptions
-                temp = temp.Skip(1).ToArray();
-                foreach(string d in temp)
-                {
-                    Console.WriteLine(d);  
-                }
-                return temp;
-
+                return File.ReadAllLines("People.csv").Skip(1).ToList();//System.Reflection.Assembly.GetExecutingAssembly().Location; current will work but it relies on unsafe assumptions
             }
         }
 
 
         // 2.
-        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows() 
-            => throw new NotImplementedException();
+        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+        {
 
+            // Define Query
+            IEnumerable<string> stateQuery = (
+                from line in CsvRows
+                let columnSeparatedLines = line.Split(',')
+                orderby columnSeparatedLines[6]
+                select columnSeparatedLines[6]
+                ).Distinct().ToList();
+
+            //Execute Query
+            IEnumerable<string> stateList = stateQuery;
+
+            return stateList;
+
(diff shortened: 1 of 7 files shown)


MarkCapka:
Working on code review for this Pull request - Mark Capka

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,105 @@
...
 {
     public class SampleData : ISampleData
     {
         // 1.
-        public IEnumerable<string> CsvRows => throw new NotImplementedException();
+        public IEnumerable<string> CsvRows // = the current return ? 2/17 lecture
+        {
+            get
+            {
+                return File.ReadAllLines(FilePath).Skip(1).ToList();//System.Reflection.Assembly.GetExecutingAssembly().Location; current will work but it relies on unsafe assumptions
```
xiokd, line 16: A possible approach when using System.Reflection.Assembly.GetExecutingAssembly().Location, would be to use it with the System.IO.GetDirectoryName method to get the absolute path of the executing assembly's directory returned as a string and then append the relative path to the file to that string. This approach however will need to be checked for null.

xiokd (commented):
Overall everything was good. There was one possible change that could be suggested.

   - Include a test that leverages a hardcoded list of Spokane-based addresses. ✔
   - Include a test that uses LINQ to verify the data is sorted correctly (do not use a hardcoded list). ✔

   - Use the `people` parameter from `ISampleData.GetUniqueListOfStates` for your data source. ✔

- Add an `IEnumberable<T> ChildItems(int maximum)` method to `Node` that returns the remaining items with a maximum number of items returned less than `maximum`.  

## Extra Credit

- Implement the homework using async/await and multi-threading by defining a new `SampleDataAsync` class that implements `IAsyncSampleData`). Refactor your `SampleData` and `SampleDataAsync` classes with minimal duplication. Be sure to refactor your tests to re-use a significant amount of the test code for both implementations. ❌✔

## Fundamentals

- Ensure you enable:
  - net6 targeted ✔
  - C# 10.0 ✔

Assignment/Assignment/SampleData.cs
A-Tanner: Thanks for the feedback, that answers a question I've been having about how to extract a directory from the commented method (which returns the path to a .dll file)

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,105 @@
...
+        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+        {
+
+            // Define Query
+            IEnumerable<string> stateQuery = (
+                from line in CsvRows
+                let columnSeparatedLines = line.Split(',')
+                orderby columnSeparatedLines[6]
+                select columnSeparatedLines[6]
+                ).Distinct().ToList();
```
mmwoodfo, line 42: I would suggest sticking with the LINQ method syntax, such as below.
```suggestion
            return CsvRows
                .Select(row => row.Split(',')[6])
                .OrderBy(state => state)
                .Distinct();
```

mmwoodfo (commented):
Good job!

mmwoodfo:
@schuywalker did you do a PR review?

PR #133, opened 2022-02-22
Assignment5+6 

mpennock5:
@Aedyn01

More test methods will be helpful.


Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,76 @@
...
 namespace Assignment
 {
     public class SampleData : ISampleData
     {
-        // 1.
-        public IEnumerable<string> CsvRows => throw new NotImplementedException();
+        private List<string>? _CsvRows;
+        private string peopleCSV = System.Reflection.Assembly.GetExecutingAssembly().Location.Replace("Assignment.dll", "People.csv");
+
+        public SampleData()
```
dcox2005, line 13: I passed the path into the constructor on the chance that the file is not located within the same location as the class file.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,76 @@
...
+            split.Sort();
+
+            return split;
+        }
+           
 
         // 3.
         public string GetAggregateSortedListOfStatesUsingCsvRows()
-            => throw new NotImplementedException();
+            => string.Join(", ", GetUniqueSortedListOfStatesGivenCsvRows());
```
dcox2005, line 45: part 3 says to create a comma separated list, this is usually done with only and "," not ", ".  I would suggest doing this without the space.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,76 @@
...
 
         // 4.
-        public IEnumerable<IPerson> People => throw new NotImplementedException();
+        public IEnumerable<IPerson> People
+        {
+            get
```
dcox2005, line 50: This is a property, I suggest creating it within the constructor.  The benefit to the way it is done here is that if the property is never called, the work for it is never done.  But if the property is called at least twice then the work is being done exponentially more.  We created a helper method that was called in the constructor and passed the results into this property.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,76 @@
...
+                                              personAttribute[3]);
+                
+                return people;
+            }
+        }
 
         // 5.
         public IEnumerable<(string FirstName, string LastName)> FilterByEmailAddress(
-            Predicate<string> filter) => throw new NotImplementedException();
+            Predicate<string> filter) =>
```
dcox2005, line 66: This looks like a more condensed and cleaner way to do this than the way we did it. Very nice

Assignment/Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,87 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assignment.Tests;
```
dcox2005, line 8: Tests names should be more meaning full. Ex "Part6_MethodReturnsUniqueList_SuccessListMatches". WhatIsBeingTested_HowItIsBeingTested_HowSuccessIsMeasured

dcox2005 (commented):
Overall, very well done.  Some changes here and there that I would suggest, and some changes for my code I will probably implement :)

Include a test that leverages a hardcoded list of Spokane-based addresses. ❌  
            I did not see a test that tested this.
Include a test that uses LINQ to verify the data is sorted correctly (do not use a hardcoded list). ✔

Consider "selecting" only the states and calling ToArray() to retrieve an array of all the state names. ❌
          ToArray was never used.
Given the array, consider using string.Join to combine the list into a single string. ✔

Sort the list by State, City, Zip. (Sort the addresses first then select). ✔
Consider using ISampleData.CsvRows in your test to verify your results. ✔

Use the people parameter from ISampleData.People property for your data source. ✔
It is recommended that, at a minimum, you use ISampleData.GetUniqueSortedListOfStatesGivenCsvRows to validate your result.

Implement IEnumerable<T> to return all the items in the "circle" of items. ✔
Add an IEnumberable<T> ChildItems(int maximum) method to Node that returns the remaining items with a maximum number of items returned less than maximum.

Assignment/Assignment/SampleData.cs
mpennock5: This is an excellent point, will definitely rework this and consider the implications of how its implemented. Good catch and thank you!

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,76 @@
...
     {
-        // 1.
-        public IEnumerable<string> CsvRows => throw new NotImplementedException();
+        private List<string>? _CsvRows;
+        private string peopleCSV = System.Reflection.Assembly.GetExecutingAssembly().Location.Replace("Assignment.dll", "People.csv");
+
+        public SampleData()
+        {
+            IEnumerable<string> CSV = File.ReadAllLines(peopleCSV);
+            CsvRows = CSV;
```
mmwoodfo, line 16:
```suggestion
            _CsvRows = CSV;

```

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,76 @@
...
+
+        }
 
         // 2.
-        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows() 
-            => throw new NotImplementedException();
+        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+        {
+            List<string> split = CsvRows.Select(x => x.Split(',')[6]).Distinct().ToList();
+            split.Sort();
```
mmwoodfo, line 37: You should use LINQs `OrderBy` for this. Additionally, you don't need to convert to a list with `ToList`.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,76 @@
...
+            { 
+                IEnumerable<Person> people = from person in CsvRows.OrderBy(state => state.Split(',')[6])
+                .ThenBy(city => city.Split(',')[5]).ThenBy(zip => zip.Split(',')[7]).ToList()
+                let personAttribute = person.Split(',')
+                                              select new Person(personAttribute[1], 
+                                              personAttribute[2], 
+                                              new Address(personAttribute[4], personAttribute[5], personAttribute[6], personAttribute[7]),
+                                              personAttribute[3]);
+                
+                return people;
```
mmwoodfo, line 60: When using LINQ, try to only use one of it's flavors (query or method flavor syntax)
The below is the solution using the method syntax (I reccommend this one as it is more C# like/looking)
```suggestion
                return CsvRows
                    .Select(x => x.Split(","))
                    .OrderBy(state => state[6])
                    .ThenBy(city => city[5])
                    .ThenBy(zip => zip[7])
                    .Select(personAttribute => new Person(
                        personAttribute[1],
                        personAttribute[2],
                        new Address(
                            personAttribute[4],
                            personAttribute[5],
                            personAttribute[6],
                            personAttribute[7]),
                        personAttribute[3])
                    );
```

PR #135, opened 2022-02-22
Assignment5+6

MarkCapka:
@TylerRose8 - partner


Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,197 @@
...
+        CircularLinkedList <IPerson> linkedList = new(People.First());
+
+        People.Skip(1).ToList().ForEach(people => linkedList.Append(people));
+        return linkedList;
+
+    }
+
+
+
+    public IEnumerable<IPerson> ChildItems(int maximum)
```
Aedyn01, line 174: Shouldn't this method be a part of the Node class?

Aedyn01 (commented):
Overall I thought your assignment came out very well. There was one "issue" that I noticed with the ChildItems method being in your SampleData class instead of the Node class. Given the instructions I assumed it meant to be a part of Node. Other than that, the only other things I can find would be "style points" or just personal preference type things. 

You have a few methods where all you do is declare a variable, then just return that variable. For example, your GetAggregateListOfStatesGivenPeopleCollection method declares states, then returns it. Instead of declaring states you can just return like "return string.Join(',', GetUniqueListOfStates(people).ToArray());" Going farther, you can also use the lambda operator and do away with the code block and return all together by simply just using "=> string.Join(',', GetUniqueListOfStates(people).ToArray());" since the method return type is a string. I doubt this is something they would deduct points from for this project, but I think it would be good to keep in mind that you can use a lambda expression for this. Theres a few methods you have where this is applicable, so rather than comment on each one of them I just made a general comment on this.

All in all, good project, just go over it one time to clean up some of the code.

kelseymcmahon (commented):

   - Include a test that leverages a hardcoded list of Spokane-based addresses. ✔
   - Include a test that uses LINQ to verify the data is sorted correctly (do not use a hardcoded list). ✔

   - Use the `people` parameter from `ISampleData.GetUniqueListOfStates` for your data source. ✔

- Implement `IEnumerable<T>` to return all the items in the "circle" of items. ❌
- Add an `IEnumberable<T> ChildItems(int maximum)` method to `Node` that returns the remaining items with a maximum number of items returned less than `maximum`. ❌

## Extra Credit

- Implement the homework using async/await and multi-threading by defining a new `SampleDataAsync` class that implements `IAsyncSampleData`). Refactor your `SampleData` and `SampleDataAsync` classes with minimal duplication. Be sure to refactor your tests to re-use a significant amount of the test code for both implementations. 

## Fundamentals

- Ensure you enable:
  - net6 targeted ✔
  - C# 10.0 ✔

**Everything looks great, my only comment would be to work on the node class file. The ChildItem(int maximum) method should be implemented in the node class.**

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,176 @@
...
-            Predicate<string> filter) => throw new NotImplementedException();
 
-        // 6.
-        public string GetAggregateListOfStatesGivenPeopleCollection(
-            IEnumerable<IPerson> people) => throw new NotImplementedException();
+    //Not sure where this function was required, but step 6 says to use it so I created it. Essentially does the same as step 2 but uses the given collection instead of getting csv rows.
+    //Might be fine to leave private.
+    private IEnumerable<string> GetUniqueListOfStates(IEnumerable<IPerson> personList)
+    {
+        IEnumerable<string> data = personList.DistinctBy(person => person.Address.State).Select(person => person.Address.State).ToList();
```
mmwoodfo, line 70:
```suggestion
        IEnumerable<string> data = personList
          .DistinctBy(person => person.Address.State)
          .Select(person => person.Address.State);
```
This is typically how long LINQ chains are formatted. Additionally, you don't need to call `ToList` to explicitly convert to a list as you are placing the result in an `IEnumerable` type as well as returning that type.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,176 @@
...
+
+    // API link for Ordering: https://docs.microsoft.com/en-us/dotnet/standard/linq/
+
+    //data in row: Id,FirstName,LastName,Email,           StreetAddress,City,State,Zip
+    //data format: 1,Priscilla,Jenyns,pjenyns0@state.gov,7884 Corry Way,Helena,MT,70577
+    public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+    {
+
+
+        IEnumerable<string> data = GetUniqueListOfStates(People).OrderBy(state => state);
```
mmwoodfo, line 94: This should be using `ISampleData.CsvRows` as its data source, not `People`.

Assignment/Assignment/SampleData.cs
mmwoodfo: Additionally, you should consider doing `select` before `distinct`. You may find it easier to work with.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,176 @@
...
+            lastName = column[2],
+            address = new Address(column[4], column[5], column[6], column[7]),
+            emailAddress = column[3],
+        }).ToList();
+        List<IPerson> newPeople = new List<IPerson>(); //People.Append didn't put anything in the list. Use an interrum list.
+        data.ForEach(field =>
+        {
+            newPeople.Add(new Person(field.firstName, field.lastName, field.address, field.emailAddress));
+        });
+        People = newPeople.OrderBy(person => person.Address.State).ThenBy(person => person.Address.City).ThenBy(person => person.Address.Zip);
```
mmwoodfo, line 61: While this works, it is very messy. You should spend some time tidying up final submissions in the future.

This might be a better approach:
```C#
        People = CsvRows
          .Select(x => x.Split(","))
          .OrderBy(x => x[6])
          .ThenBy(x => x[5])
          .ThenBy(x => x[7])
          .Select(person =>
          new Person(
              person[1],
              person[2],
              new Address(
                  person[4],
                  person[5],
                  person[6],
                  person[7]
              ),
              person[3]));
```

Although if you take this approach, you may consider putting this directly on the getter of the People property rather than setting it in the constructor. Take a look at the Assignment 5+6 solution for another acceptable way to do this via the constructor.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,176 @@
...
+    }
+
+    // 6.Implement ISampleData.GetAggregateListOfStatesGivenPeopleCollection(IEnumerable<IPerson> people) to return a string that contains a unique, comma separated list of states. ❌✔
+    // Use the people parameter from ISampleData.GetUniqueListOfStates for your data source. ✔
+    //TODO: At a minimum, use System.Linq.Enumerable.Aggregate LINQ method to create your result. ✔
+    //TODO: Don't forget the list should be unique. ✔
+    //TODO: It is recommended that, at a minimum, you use ISampleData.GetUniqueSortedListOfStatesGivenCsvRows to validate your result.
+    public string GetAggregateListOfStatesGivenPeopleCollection(IEnumerable<IPerson> people) =>
+  
+     people.Select(x=>x.Address.State).Distinct().Aggregate((states, nextState) => states + ", " + nextState);
```
mmwoodfo, line 145: Good job selecting before calling `Distinct`

Assignment/Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,137 @@
...
+        ISampleData sampleData = new SampleData();
+        IEnumerable<(string FirstName, string LastName)>? aEmails = sampleData.FilterByEmailAddress(new Predicate<string>(email => email.StartsWith('a')));
+        Assert.AreEqual<int>(4, aEmails.Count());
+        Assert.AreEqual<(string FirstName, string LastName)>(("Arthur", "Myles"), aEmails.First());
+    }
+
+
+
+    //6. GetAggregateListOfStatesGivenPeopleCollection: 
+    //TODO: It is recommended that, at a minimum, you use ISampleData.GetUniqueSortedListOfStatesGivenCsvRows to validate your result.
```
mmwoodfo, line 91: Looks like this is still a TODO?

Assignment/Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,137 @@
...
+    //TODO from part 2 
+    //TODO: Include a test that leverages a hardcoded list of Spokane-based addresses. ✔           //TODO drop question if can't figure out from recording, what his intention is. ALl addresses will be Washington. (so should probably sort on the other values like state, city, zip etc?
+    //Include a test that uses LINQ to verify the data is sorted correctly (do not use a hardcoded list). ✔
+    // Include a test that leverages a hardcoded list of Spokane-based addresses. 
+
+    [TestMethod]        //TODO come back to once we figure out the hardcoded, washington state stuff
+    public void GetUniqueSortedListOfStatesGivenCsvRowsTest_HardCodedInput_Success()
+    {
+        ISampleData sampleData = new SampleData();
+        IEnumerable<string> csvRows = sampleData.CsvRows; //holds each of our individual values in the row
```
mmwoodfo, line 46: So when it says 'Hardcoded list' it means something like this:
```C#
List<string> hardCodedListExample = new(){
"1,Priscilla,Jenyns,pjenyns0@state.gov,7884 Corry Way,Spokane,WA,70577"
"2,Karin,Joder,kjoder1@quantcast.com,03594 Florence Park,Spokane,WA,71961"
"3,Chadd,Stennine,cstennine2@wired.com,94148 Kings Terrace,Spokane,WA,59721"
}
```

PR #139, opened 2022-02-22
Assignment5+6 @Disciple4GIT

nwnel-017:


NazariyPogo:

   - Include a test that leverages a hardcoded list of Spokane-based addresses. ❌ No hardcoded list visible.
   - Include a test that uses LINQ to verify the data is sorted correctly (do not use a hardcoded list). ❌ Method started but not finished.

   - Consider "selecting" only the states and calling `ToArray()` to retrieve an array of all the state names. ✔ Did not use ToArray() but works fine.
   - Given the array, consider using `string.Join` to combine the list into a single string. ✔ Used aggregate.

   - Sort the list by State, City, Zip. (Sort the addresses first then select). ❌ Does not look sorted

   - Use `ISampleData.People` for your data source. ❌

   - Use the `people` parameter from `ISampleData.GetUniqueListOfStates` for your data source. ❌ Not seeing the data from GetUniqueListOfStates used.

- Add an `IEnumberable<T> ChildItems(int maximum)` method to `Node` that returns the remaining items with a maximum number of items returned less than `maximum`.  ❌ Not yet implemented

## Extra Credit

- Implement the homework using async/await and multi-threading by defining a new `SampleDataAsync` class that implements `IAsyncSampleData`). Refactor your `SampleData` and `SampleDataAsync` classes with minimal duplication. Be sure to refactor your tests to re-use a significant amount of the test code for both implementations. ❌

## Fundamentals

- Ensure you enable:
  - net6 targeted ✔
  - C# 10.0 ❌




Looks on the right track, just a bit more.

NicholasJohnsonDev:
Starting review. 
I know it already has a review but every other pull request already has a review too and I see that they have made new commits since the pull review.

NicholasJohnsonDev:
As this has already been reviewed I am going to try to address things that were not covered, give countering opinions, but also look closely at your newest commits.

C# 10.0 ✔ and enabled .NET analyzers for both projects ✔
I disagree with the previous reviewer's stating you forgot these two. This was actually given to us already done for the assignment. Mark had been typically having us add these in the csproj files but this time he gave a Directory.Build.props that works not at the project level but the solution level and the projects inherit from it.
https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022
And though you never explicitly had <LangVersion>10.0</LangVersion> the default (when not specified otherwise) is to use the most recent latest major version of C#, which is 10. I would recommend adding <ImplicitUsings>enable</ImplicitUsings> and <AnalysisMode>AllEnabledByDefault</AnalysisMode> to help auto-import usings and show more warnings, respectively.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

The Nuget Packages you are referencing are out of date, 4 total. 3 are minor but Microsoft.NET.Test.Sdk is a major version behind. (I know Mark gave it to us out of date but it is good practice to do.) 
You also are not utilizing the IntelliTectAnalyzer, which has helped me notice issues I would have otherwise missed. (But other warnings don't apply to our project and just need to be suppressed, so it isn't perfect.)
https://www.nuget.org/packages/IntelliTect.Analyzers/

I had to change this, but I don't know if this is something that resets on a new clone.
Change the "Copy to" property on People.csv to "Copy if newer" so that the file is deployed along with your test project. ❌

Your tests failed for me but only because the FilePath given does not work (probably just an operating system issue). Assignment Tests line: 12.
First, I would like to recommend you use the @ which is the string verbatim suffix, this will allow the string to be more human-readable. In this instance, you will not need to use additional slashes as the escape character.
Before: public string FilePath = "N:\\Assignment5+6\\Assignment\\Assignment\\People.csv";
After: public string FilePath = @"N:\Assignment5+6\Assignment\Assignment\People.csv";
But even better, you should use AppDomain.CurrentDomain.BaseDirectory + "People.csv" as Mark suggested in class.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/verbatim

After changing the file path to work for me, your tests that run passed, with the exception of InitializeSampleDataClass_WithEmptyFile_Failure(), but you have a comment to expect it to. For the tests that don't run, it seems like it is trying and then gives up, perhaps there is a time limit for tests? I would definitely ask the professor tomorrow about this.

You seem like your making it harder on yourself in SampleData when you keep going back and forth from a list.

Though where you are at has a lot of incompletes, you are really close to getting all of them done.

nwnel-017:
Ok thank you so much. I believe it is updated is there any way you could double check to see if my pull request is updated, I feel unsure if I did it the right way. The latest commit should be 'project is finished'
________________________________

mmwoodfo:
It looks like it worked to me. :) I see the commits from your other PR here.

nwnel-017:
Perfect, thank you again!
________________________________

ConsoleApp1/ConsoleApp1/Program.cs
```diff
@@ -0,0 +1,4 @@
+﻿// See https://aka.ms/new-console-template for more information
+Console.WriteLine("Hello, World!!!");
+
+//This is my change
```
mmwoodfo, line 4: This shouldn't be included in your PR

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
-        public string GetAggregateSortedListOfStatesUsingCsvRows()
-            => throw new NotImplementedException();
+            if (!File.Exists(filePath))//case if file does not exist
+            {
+                throw new FileNotFoundException("Error, file does not exist");
+            }
+            if (new FileInfo(filePath).Length == 0)//case if file passed in is empty
+            {
+                throw new ArgumentNullException("Error, file is empty");
+            }
```
mmwoodfo, line 22: You should be setting the .csv file properties to be `copy if newer`. Currently, it is set to `do not copy`
image

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
 
-        // 4.
-        public IEnumerable<IPerson> People => throw new NotImplementedException();
+            try
+            {
+                _CsvRows = File.ReadAllLines(filePath).Skip(1).ToList();
+            }catch (Exception)
```
mmwoodfo, line 27: You shouldn't catch just any exception and then rethrow that exception as an ArgumentNullException.....
What if the exception was that the file is corrupt? An ArgumentNullException being thrown would not make any sense. If you are just going to catch and throw, you might as well not catch at all and let the original exception occur.

The only time an ArgumentNullException should be thrown is if the parameter `string filePath` is null.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
 
-        // 6.
+        // 2.->Finished
+        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+        {
+            List<string> states = new List<string>();
+            
+            states = CsvRows.Select(item => item.Split(",")).OrderBy(item => item[6]).Select(item => item[6]).Distinct().ToList();
+            states.Zip(states, (current, next) => string.Compare(current, next) <= 0).All(x => x);
+            return states;
```
mmwoodfo, line 48: You have a lot of additional steps taking place that don't really need to be happening.
You should be able to solve the problem by simply doing the following:
```suggestion
            var states = CsvRows
                .Select(item => item.Split(",")[6])
                .OrderBy(item => item)
                .Distinct();
            return states;
```
You could even just do the return on one line if you wanted.

But it looks like you have two selects, you just need one select at the top and you do not need to convert to a list. You also do not need to use `.Zip`

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
+            return statesString;
+        }
+
+        // 4.->Finished
+        public IEnumerable<IPerson> People { get {
+                IEnumerable<IPerson> people =
+                    CsvRows.Select(line => line.Split(",")).OrderBy(line => line[5]).ThenBy(line => line[6]).ThenBy(line => line[7])
+                    .Select(item => new Person(item[1], item[2], new Address(item[4], item[5], item[6], item[7]), item[3])).ToList();
+                return people;
+            }
```
mmwoodfo, line 66:
```suggestion
        public IEnumerable<IPerson> People { get => CsvRows
                .Select(line => line.Split(","))
                .OrderBy(line => line[5])
                .ThenBy(line => line[6])
                .ThenBy(line => line[7])
                .Select(item => new Person(
                    item[1],
                    item[2],
                    new Address(
                        item[4],
                        item[5],
                        item[6],
                        item[7]),
                    item[3]));
```
You should consider writing long LINQ chains on multiple lines for better readability.

Additionally, you are sorting by city then state then zip. This should sort by State then city then zip.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
+            
+        }
+        // 5. -> Finished?
+        public IEnumerable<(string FirstName, string LastName)> FilterByEmailAddress( 
+            Predicate<string> filter)
+        {
+            List<IPerson> people = People.ToList();
```
mmwoodfo, line 73:
```suggestion
```
You don't need this here.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
+            IEnumerable<(string FirstName, string LastName)> results = 
+                People.ToList().FindAll(person => filter(person.EmailAddress)).
+                Select(person => (person.FirstName, person.LastName));
```
mmwoodfo, line 76:
```suggestion
            IEnumerable<(string FirstName, string LastName)> results = People
                    .Where(person => filter(person.EmailAddress))
                    .Select(person => (person.FirstName, person.LastName));
```
You don't need to convert `People` to a list if you use LINQs `Where`. `FindAll` is not a LINQ operation, but a list operation.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
+        }
+
+        // 6.->Finished
         public string GetAggregateListOfStatesGivenPeopleCollection(
-            IEnumerable<IPerson> people) => throw new NotImplementedException();
+            IEnumerable<IPerson> people)
+        {
+            if(people == null)
+            {
+                throw new ArgumentNullException("error, bad people list");
```
mmwoodfo, line 87: A more descriptive error message would be better. For example, we already know it's an exception so no need to write `error`.  Additionally, `bad` is not very descriptive. Maybe something along the lines of `"People list cannot be null"` or `"People list is null"`

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
+            }
+
+            IEnumerable<string> statesList = people.Select(person => (person.Address.State)).Distinct().ToList();
```
mmwoodfo, line 90: Again, no need to convert to a list.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
+            return statesList.Aggregate((state1, state2) => state1 + ", " + state2);
```
mmwoodfo, line 91: 👍🏻

Assignment/Assignment/SampleData.cs
mmwoodfo, line 76: By the way - Good job with your names for the LINQ selector variables... Most people will name this `x` or something, but naming it `people` or even just `p` is alot cleaner so good job with that. 👍🏻 
-----------&#8595;
Select(person => (person.FirstName, person.LastName));

Assignment/Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,182 @@
...
+    [TestClass]
+    public class SampleDataTests
+    {
+        public string FilePath = @"N:\\Assignment5+6\\Assignment\\Assignment\\People.csv";//I understand this isn't the right way
+        //because the filepath on your machine is different--> i tried using the statement below but it crashed on my system
+        //every time--> my machine was being unpredictable and doing different things every time i ran the program, i think it was
+        //a problem with the computer
+
+
+        //public string FilePath = AppDomain.CurrentDomain.BaseDirectory + "People.csv";//--> Why isnt this working
```
mmwoodfo, line 21: This isn't working because you have to change the properties on your `People.csv` to be `Copy if Newer` (see comment in SampleData.cs w/ screenshot about this.)

In the screenshot below, you see that I've changed that property on the `People.csv` to be `Copy if Newer` and have uncommented line 21 with everything working.

image

mmwoodfo:
@nwnel-017 @Disciple4GIT
Hey guys, Mark mentioned that one of you asked about the low score for this assignment. I wanted to let you guys know that you both were missing an additional 20 points that did not get added on due to an error on my part, so you should see your grades updated in canvas. If you would like a further breakdown of your grade please email EWUInstructors@IntelliTect.com.

Thank you! Sorry about the grading error.

nwnel-017:
No problem! Thanks so much for double checking!!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #140, opened 2022-02-22
Assignment5+6 @dcox2005

hgill17:
Assignment5+6 
Partner @dcox2005 David Cox


Assignment/Assignment.Tests/NodeTests.cs
```diff
@@ -0,0 +1,233 @@
...
+    {
+        Node<string> myNode = new(null!);
+        Assert.IsNotNull(myNode);
+    }
+
+    [TestMethod]
+    public void ToString_ValueShouldBeReturned_Success()
+    {
+        string value = "Some Value";
+        Node<string> myNode = new(value);
```
whitneybolar, line 44: Inside these ToString methods, you can test to see if you are getting a string: Assert.IsInstanceOfType(..,..)

Assignment/Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,233 @@
...
+namespace Assignment.Tests;
+
+[TestClass]
+public class SampleDataTests
+{
+    [DeploymentItem("People.csv")]
+    [TestMethod]
+    public void GetPath_VerifyCSVExistsInPath_Success()
+    {
+        Assert.IsTrue(File.Exists("People.csv"));
```
whitneybolar, line 17: That's a good idea! I did not think to verify the existence in the beginning.

whitneybolar (commented):

Include a test that leverages a hardcoded list of Spokane-based addresses. ❌ I could be mistaken, but I do not clearly see any Spokane-based addresses. double check 
Include a test that uses LINQ to verify the data is sorted correctly (do not use a hardcoded list). ✔

Consider "selecting" only the states and calling ToArray() to retrieve an array of all the state names. ✔
Given the array, consider using string.Join to combine the list into a single string. ✔

Sort the list by State, City, Zip. (Sort the addresses first then select). ❌ the sort was not on sampleData.cs. double check
Consider using ISampleData.CsvRows in your test to verify your results. ✔

Use the people parameter from ISampleData.GetUniqueListOfStates for your data source. ✔
It is recommended that, at a minimum, you use ISampleData.GetUniqueSortedListOfStatesGivenCsvRows to validate your result.

Implement IEnumerable<T> to return all the items in the "circle" of items. ✔
Add an IEnumberable<T> ChildItems(int maximum) method to Node that returns the remaining items with a maximum number of items returned less than maximum.

nwnel-017:
Looks like every other pull request has been reviewed for this assignment. There are not any new commits since the last review. Looking through your code, this looks like a mostly completed assignment. You still need to implement the hard coded test with spokane addresses for part 2, and sort the list by state in part 4 where you implement the People property in SampleData. Besides this, your project looks very good.

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,100 @@
...
         // 2.
-        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows() 
-            => throw new NotImplementedException();
+        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+        {
+            List<string> states = CsvRows.
+                Select(person => person.Split(',')[6]).
+                Distinct().
+                OrderBy(state => state).
+                ToList();
```
mmwoodfo, line 21:
```suggestion
            List<string> states = CsvRows
                .Select(person => person.Split(',')[6])
                .Distinct()
                .OrderBy(state => state)
                .ToList();
```
Good question! The `.` typically goes at the beginning of a new LINQ statement. This is because it makes it easy remove/rearrange/add statements without touching the lines above/below too much. Also a bit easier to see the `.` when it goes at the start of the statement.

mmwoodfo: Additionally, you don't need to convert this to a list here. In some cases, it's actually better to not do so - https://www.tutorialsteacher.com/linq/linq-deferred-execution

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,100 @@
...
+            return string.Join(",", states);
+        }
+
+        // 7
+        //Check Node.cs
+       
+        public SampleData(string path)
+        {
+            CsvRows = CSVParser(path);
+            People = SetPeople();
```
mmwoodfo, line 70: I think you guys would find the way Grant solve this assignment w/ a named dictionary interesting
https://github.com/IntelliTect-Samples/EWU-CSCD371-2022-Winter/blob/Assignment5%2B6Complete/Assignment/Assignment/SampleData.cs

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,100 @@
...
 
         // 5.
-        public IEnumerable<(string FirstName, string LastName)> FilterByEmailAddress(
-            Predicate<string> filter) => throw new NotImplementedException();
+        public IEnumerable<(string FirstName, string LastName)> FilterByEmailAddress(Predicate<string> filter)
+        {
+            List<IPerson> persons = People.ToList();
+            IEnumerable<(string FirstName, string LastName)> results = People.ToList().
+                FindAll(person => filter(person.EmailAddress)).
+                Select(person => (person.FirstName, person.LastName));
```
mmwoodfo, line 44: You don't need to convert this to a list if you use LINQ operations (`FindAll` is a list operation, hence why it only works on a list object)

Consider the following solution:
```suggestion
            IEnumerable<(string FirstName, string LastName)> results = People
                .Where(person => filter(person.EmailAddress))
                .Select(person => (person.FirstName, person.LastName));
```

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,100 @@
...
+        {
+            List<string> states = people
+                .Select(person => (person.Address.State))
+                .Aggregate((states, next) => $"{states},{next}")
+                .Split(",")
+                .Distinct()
+                .OrderBy(state => state)
+                .ToList();
+
+            return string.Join(",", states);
```
mmwoodfo, line 61:
```suggestion
            return people
                .Select(person => (person.Address.State))
                .Distinct()
                .OrderBy(state => state)
                .Aggregate((states, next) => $"{states},{next}");
```
Instead of aggregating, then splitting, then joining... Simply do all of the selections/ordering/distinct calls then call Aggregate.

PR #141, opened 2022-02-22
Assignment5+6

GavenElder:
implemented 1-6.


DakotaCondos (commented):
Overall looks good, but where are the unit tests?

1


2
Include a test that leverages a hardcoded list of Spokane-based addresses. ❌
Include a test that uses LINQ to verify the data is sorted correctly (do not use a hardcoded list). ❌


3
Consider "selecting" only the states and calling ToArray() to retrieve an array of all the state names. ❌ (Not sure how to grade something thats a consideration.. You are using a list and not an array to return the same values. Also there is a space between your values here which may be problematic if not trimmed later on)
Given the array, consider using string.Join to combine the list into a single string. ✔ (same as above comment, you are using Aggregate and not string.Join to achieve the same result)

here is a sample of how our group did this portion:
string[] state = GetUniqueSortedListOfStatesGivenCsvRows().Select(x => x).ToArray();
	return string.Join(",", state);


4
Sort the list by State, City, Zip. (Sort the addresses first then select). ❌ (this is not implemented)
Consider using ISampleData.CsvRows in your test to verify your results. ❌ (No visible unit tests)


5


6
Use the people parameter from ISampleData.People property for your data source. ✔
It is recommended that, at a minimum, you use ISampleData.GetUniqueSortedListOfStatesGivenCsvRows to validate your result. (No visible unit tests)


7
Implement IEnumerable<T> to return all the items in the "circle" of items. ❌
Add an IEnumberable<T> ChildItems(int maximum) method to Node that returns the remaining items with a maximum number of items returned less than maximum. (Not yet implemented)

mmwoodfo:
@GavenElder This code is late & you don't have a partner for the assignment.

PR #148, opened 2022-03-08
Assignment7+8

kelseymcmahon:
Partner: @whitneybolar

We were having some difficulties understanding catching errors with AXIOS and for this bullet:
"If the joke API returns an error display a message to try again in a few moments". 
Any feedback would be great.


A-Tanner:
I will review this

Assigment/js/index.js
```diff
@@ -0,0 +1,74 @@
...
+                    console.log(error.response.headers);
+                } else if (error.request) {
+                    // The request was made but no response was received
+                    console.log(error.request);
+                } else {
+                    let jokeSetup = document.querySelector(".jokeSetup");
+                    jokeSetup.innerHTML = "<p>Error getting joke, fetching another joke...</p>";
+                    console.log('Error', error.message);
+                    sendGetRequest();
+                }
```
A-Tanner, line 22: Given the logical structure, an error may not always be printed to end user (i.e. on the actual webpage.) Consider putting the final "else" block into a "finally" block if you want to keep console diagnostics, or making it the only code that runs if axios produces an error. Either response would be sufficient to display the error message every time axios returns an error.

Assigment/css/site.css
```diff
@@ -0,0 +1,238 @@
...
+    }
+}
+
+
+/* Mobile CSS Styles */
+@media screen and (max-width: 650px) {
+    body {
+        width: 100%;
+    }
+
```
A-Tanner, line 225: I can't find a case (including default body width %) where 50% width is given to body.

Assigment/js/index.js
```diff
@@ -0,0 +1,74 @@
...
+        else {
+            jokeSetup.innerHTML = "<p>" + response.data.setup + "</p>";
+            setTimeout(() => { jokePunchline.innerHTML = "<p>" + response.data.delivery + "</p>"; }, 4000);
+        }
+    }
+    catch (error) {
+        console.error(error);
+        jokeSetup.innerHTML = "<p>Error getting joke, fetching another joke...</p>";
+        sendGetRequest();
+    }
```
A-Tanner, line 43: It's hard to tell if this catch will ever run, because in the code preceding it there is a catchall for any potential axios error. I want to say that this would only run if there was an error outside of the Axios.Get method (but still originating from sendGetRequest), but I'm not qualified enough to guarantee that statement is true

A-Tanner (changes requested):
The project should look like the image below ✔
The project should have an index.html file ✔
With an index.js file in a js folder ✔
And a site.css file in a css folder (along with other css files) ✔
There should be a <head>
With a favicon ✔
And a title ✔
The <body> should use the middle 70% of the page ✔
Menu Bar
The page should have a title/menu bar ✔
Use Font Awesome for the hamburger icon (https://fontawesome.com/start[](https://github.com/IntelliTect-Samples/EWU-CSCD371-2022-Winter/tree/Assignment7+8#create-a-card-with-divs-and-css)[](https://github.com/IntelliTect-Samples/EWU-CSCD371-2022-Winter/tree/Assignment7+8#show-jokes-on-the-card)) ✔
When clicked, the menu should appear under it ✔
The menu should have four buttons with creative names (the buttons don't need to do anything) ✔
Create a card with <div>s and CSS
Create a card from <div>s ✔
The card should have a header, body, and footer ❌✔ --> There are Header/Body/Footer tags that can be used in place of <div>. I'm unsure how important this is and have submitted a question to the teams chat.
The CSS for this should be in the CSS file and work easily for another card. ✔
Show jokes on the card
Use the joke API from the first assignment to display a joke on a card (https://v2.jokeapi.dev/joke/Programming) ✔
Use Axios from the CDN for the API call. (https://cdn.jsdelivr.net/npm/axios/dist/axios.min.js[](https://github.com/IntelliTect-Samples/EWU-CSCD371-2022-Winter/tree/Assignment7+8#extra-credit))
If the joke API returns an error display a message to try again in a few moments ❌-->While an error in the request _may_ trigger the else clause to display the error message, due to your current logic structure, an error is not guaranteed to display an error message outside of console.
The punch line should not show immediately, but appear after 4 seconds ✔
There should be a button to get another joke ✔
There should be <a> tag that links to the joke api documentation page ❌--> It currently links to the response page, instead of the joke api documentation. Proper link to the API documentation : https://v2.jokeapi.dev
This <a> should be manually styled as a button ✔
Buttons should have a background that is a gradient ✔
All buttons (even the <a> that you made into a button) should have a hover effect ✔
Extra Credit
Do the following:

Make the buttons in the menu do something creative ✔
Rather than just making the menu appear, have it fade in and slide down. ✔
Have the page jump to 100% width on a smaller screen and 50% on a larger screen with 70% on a regular screen. ❌ --> Scales to 100% with and 70% width, but I cannot find a case that would allow body to scale to 50% width

Great job so far! Only very minor errors. I had a couple of laughs from your creative buttons.

mmwoodfo (commented):
This all looks great! Good job with the extra credit

Assigment/js/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+            if (response.data.type == "single") {
+                jokeSetup.innerHTML = "<p>" + response.data.joke + "</p>";
+            }
+            else {
+                jokeSetup.innerHTML = "<p>" + response.data.setup + "</p>";
+                setTimeout(() => { jokePunchline.innerHTML = "<p>" + response.data.delivery + "</p>"; }, 4000);
+            }
+        })
+        .catch(function (error) {
```
mmwoodfo, line 26: What exactly were you struggling with here? This looks good to me.

PR #157, opened 2022-03-08
Assignment7+8

MarkCapka:
@TylerRose8 is my partner

Mostly good to go, a few outstanding formatting, css corrections, and slight tweaks, but overall, looking good!


tremorris1999:
I will review

schuywalker:
I will review

Assignment7+8/index.html
```diff
@@ -0,0 +1,71 @@
...
+    <link rel="stylesheet" href="css/site.css">
+    <link rel="stylesheet" href="css/all.css">
+</head>
+
+<body class="centered two-thirds-width">
+
+    <div class="page-header">
+        <h1 class="eagle-red">
+            <span>EWU</span>
+            <span style="color:white" class="float-right" id="menu-button">Menu <i class="fa-solid fa-bars"></i></span>
```
tremorris1999, line 17: no inline CSS
```suggestion
            <span class="float-right" id="menu-button">Menu <i class="fa-solid fa-bars"></i></span>
```

Assignment7+8/css/site.css
```diff
@@ -0,0 +1,135 @@
...
+    padding: 1em;
+}
+
+button {
+    background-color: lightgrey;
+    color: black;
+    padding: 14px 20px;
+    font-weight: bold;
+    font-size: 1.1em;
+    border: linear-gradient(to bottom right, lightGrey 0%, black 100%);
```
tremorris1999, line 58: Validator: 

> Value Error : border linear-gradient(to bottom right, lightgrey 0%, black 100%) is not a border value : linear-gradient(to bottom right, lightgrey 0%, black 100%)

Assignment7+8/css/site.css
```diff
@@ -0,0 +1,135 @@
...
+}
+
+.get-joke {
+    background-color: #797979;
+    background:linear-gradient(to bottom right, rgb(136, 177, 179) 0%, rgb(24, 151, 224) 100%);
+    color: white;
+    font-weight: 100;
+    padding: 14px 20px;
+    border-radius: 0;
+    border: linear-gradient(to bottom right, grey 0%, black 100%);
```
tremorris1999, line 105: Validator: 

> Value Error : border linear-gradient(to bottom right, grey 0%, black 100%) is not a border value : linear-gradient(to bottom right, grey 0%, black 100%)

Assignment7+8/css/site.css
```diff
@@ -0,0 +1,135 @@
...
+ }
+
+.joke-api:hover{
+  /* background-color: #B7142E;  */
+  background:linear-gradient(to right, #000000 0%, #B7142E 100%);
+  color: white;
+}
+
+@font-face{
+    font-family: 'Times New Roman', Times, serif;
```
tremorris1999, line 132: Validator: 

> Value Error : font-family Too many values or values are not recognized : 'Times New Roman',Times,serif

Assignment7+8/css/site.css
```diff
@@ -0,0 +1,135 @@
...
+    src: "js/index.js";
```
tremorris1999, line 133: Validator: 

> "js/index.js" is not a src value : "js/index.js"

Assignment7+8/css/site.css
```diff
@@ -0,0 +1,135 @@
+.centered{
+    margin: auto;
+}
+
+.two-thirds-width{
+    width: 66.6%;
```
tremorris1999, line 6: Should be 70%

Assignment7+8/index.html
```diff
@@ -0,0 +1,71 @@
...
+                <p id="setup"></p>
+            </div>
+            <div>
+                <p id="delivery"></p>
+            </div>
+        </div>
+        <div class="card-footer">
+           
+                <button id="get-joke" class = "get-joke">Another Joke</button>
+                <a id="joke-api" class = "joke-api" >Joke API</a>
```
tremorris1999, line 62:
```suggestion
                <a id="joke-api" class = "joke-api" href="https://v2.jokeapi.dev/">Joke API</a>
```

Assignment7+8/js/index.js
```diff
@@ -0,0 +1,85 @@
...
+//NOTE: TODO: you can call getJoke instead for immediate output I think or jokeOutput
+//above could also potentially be more like the other example from class: 
+/*
+        from class Lecture15-web
+        console.log(new Joke().getJoke());
+        setInterval(function() {console.log(new Joke().getJoke();)}, 4000); //may also wait 4 seconds, however, one of these will cause an issue iirc 
+
+*/
+
+function apiLink() {
```
tremorris1999, line 12: No need for this function if you use the href I added in *index.html*

Assignment7+8/js/index.js
```diff
@@ -0,0 +1,85 @@
+var jokeOutput;
```
tremorris1999, line 1: You could just combine the two *getJoke* and *tellJoke* functions and not need the global.

Assignment7+8/js/index.js
```diff
@@ -0,0 +1,85 @@
...
+        delivery.innerText = ""; 
+    }
+    setTimeout(getJoke, 400);
+}
+
+function toggleMenu() {
+    document.getElementById("drop-down-menu").hidden = !document.getElementById("drop-down-menu").hidden;
+}
+
+document.getElementById("menu-button").addEventListener("click", toggleMenu);
```
tremorris1999, line 82: https://www.w3schools.com/jsref/event_onclick.asp

tremorris1999 (changes requested):
## Instructions

### General
- The project should look like the image below ✔
- The project should have an index.html file ✔
  - With an index.js file in a js folder ✔
  - And a site.css file in a css folder (along with other css files) ✔
- There should be a `<head>`
  - With a favicon ✔
  - And a title ✔
- The `<body>` should use the middle 70% of the page ❌ (66.6% != 70%)

### Menu Bar
- The page should have a title/menu bar ✔
  - Use Font Awesome for the hamburger icon (https://fontawesome.com/start) ✔ 
  - When clicked, the menu should appear under it ✔
  - The menu should have four buttons with creative names (the buttons don't need to do anything) ✔

### Create a card with `<div>`s and CSS
- Create a card from `<div>`s ✔
  - The card should have a header, body, and footer ✔
  - The CSS for this should be in the CSS file and work easily for another card. ✔

### Show jokes on the card
- Use the joke API from the first assignment to display a joke on a card (https://v2.jokeapi.dev/joke/Programming) ✔
  - Use Axios from the CDN for the API call. (https://cdn.jsdelivr.net/npm/axios/dist/axios.min.js)
  - If the joke API returns an error display a message to try again in a few moments ✔
  - The punch line should not show immediately, but appear after 4 seconds ❌
  - There should be a button to get another joke ✔
  - There should be `<a>` tag that links to the joke api documentation page ✔
    - This `<a>` should be manually styled as a button ✔
- Buttons should have a background that is a gradient ✔
- All buttons (even the `<a>` that you made into a button) should have a hover effect  ❌ (missing menu buttons)

## Extra Credit
Do the following:

1. Make the buttons in the menu do something creative ❌
2. Rather than just making the menu appear, have it fade in and slide down. ❌
3. Have the page jump to 100% width on a smaller screen and 50% on a larger screen with 70% on a regular screen. ❌

## The page should look like this
Page to Replicate

## Fundamentals
- Ensure that:
  - This is built with raw HTML, CSS, and JavaScript, no external libraries aside from Axios and a font library ✔
  - The index.html file should work when opened from a file explorer ✔
  - Keep all files in appropriate folders ✔
  - Have warning free HTML, CSS, and JavaScript ❌
  - Use as little css as possible ✔
  - No inline CSS ❌
  - Validate HTML with: https://validator.w3.org/nu/#textarea (or similar) ✔
  - Validate CSS with: https://jigsaw.w3.org/css-validator/#validate_by_input (or similar) ❌ (<html><body>


58 |   | Value Error :  border                                  linear-gradient(to bottom right, lightgrey 0%, black 100%) is not a border value :                                                                       linear-gradient(to bottom right, lightgrey 0%, black 100%)
-- | -- | --
105 |   | Value Error :  border                                  linear-gradient(to bottom right, grey 0%, black 100%) is not a border value :                                                                       linear-gradient(to bottom right, grey 0%, black 100%)
132 |   | Value Error :  font-family                                  Too many values or values are not recognized :                                                                       'Times New Roman',Times,serif
133 |   | "js/index.js" is not a src value :                                                                       "js/index.js"


</body>
</html>)
  - No errors when loading the page ✔
  - Choose simplicity over complexity ❌ (scripting was messy and had unnecessary globals and split functions)

Overall I think this could use a lot of refinement. I imagine you're having some corrections to push before the deadline. Good work for the most part though!

schuywalker:
Ok I think the 1800 changed files is causing the page to become unresponsive when I try to review the code. I tried multiple browsers and restarting my computer, so I'm just going to leave my comments here

schuywalker:
General
The project should look like the image below ✔ #MOSTLY
The project should have an index.html file ✔
With an index.js file in a js folder ✔
And a site.css file in a css folder (along with other css files) ✔
There should be a <head>
With a favicon ✔ #will update before Thurs 
And a title ✔
The <body> should use the middle 70% of the page ❌✔ #confirm quantity - but i think so 

SA: You must have a good reason for using 66.6 - might be good to comment so you don’t get marked down? I’m a little puzzled by the choice to make ‘centered’ and ‘two-thirds-width’ their own classes. You have two classes with a single styling in them - this seems like inline styling, but with more lines of code. Multiple CSS classes that do specific things, but both of these are used only once. I’d advise keeping classes broader and more often than not, named after the actual element that they’re styling (i.e. class=”content-body”).

schuywalker:
Menu Bar
The page should have a title/menu bar ✔
Use Font Awesome for the hamburger icon (https://fontawesome.com/start) ✔
When clicked, the menu should appear under it ✔
The menu should have four buttons with creative names (the buttons don't need to do anything) ✔

SA: Ok thoughts on how to pad the menu. Tl;dr is 1) remove eagle red padding and put that padding in the page-header 2) remove height from page-header (not completely sure why that fixes the vertical centering). **Classes should describe the object on the webpage, not the styling that the class applies. Favor classes over id’s.** 


Ok so we have the container which is page-header, and then 3 objects inside. The padding for the objects inside the black box container should be affected by the containers’ padding (right now eagle-red’s padding is affecting). Eagle-red is the name of the color - the class name should be something like ‘EWU-text.’ Be sure to remove the inline span style on “float-right” so you don’t lose points - this can go in the styles for the id. Choose the id of menu button or the class-name of float-right. My choice would be class=”menu-button” because you may have other menu-buttons where you want to reuse this styling. 

Same thing for the table - no reason why this shouldnt have just one class or one id. Classes should describe the object on the webpage, not the styling that the class applies.

On the topic of id’s, I think some of your id’s should be class’s/ className’s. It’s generally recommended to avoid id’s when possible, some IDE’s consider them deprecated => https://dev.to/clairecodes/reasons-not-to-use-ids-in-css-4ni4 Another example from menu-button is id=”drop-down-menu”. This is something you might want to reuse in multiple places throughout your webpage (obviously not applicable to this assignment but just thinking of design principles). Hopefully this menu would be the same component on all the webpages, so no difference there, but it could be nice to have this styling available on other places on the website? 

Back to the header styling: as far as I can tell, a professional web developer would use flexbox here (display: flex). Obviously we didn’t go over flexbox because they have a learning curve, so using float in this project makes sense. But centering/ organizing things on the page was notoriously difficult before flexboxs, and as far as I can tell people don’t really use float anymore.

I didn’t like the span’s within the h1 at first, but I’m starting to like it more way to do it (without flexbox). Just seems weird to put it all inside an H1 since you have to deal with all the default styling.

More small things… may or may not want to have table stylings apply to every single table throughout your website, so I’d think a className would be better? Tables could be used for a lot of things. Same thing as above: get rid of the id’s and make them a single class. Probably can get rid of the p tags too, but if its working just leave it.

schuywalker:
### Create a card with divs and CSS
Create a card from divs ✔
The card should have a header, body, and footer ✔
The CSS for this should be in the CSS file and work easily for another card. ✔ SA: great names cant wait to Find Miracle Max
Show jokes on the card
Use the joke API from the first assignment to display a joke on a card (https://v2.jokeapi.dev/joke/Programming) ✔ #note: have to click button for first joke to appear `**Nice thinking! Could solve by pulling joke up immediately or also by changing the joke retrieval button to be “Get Joke” and then change to “Another Joke” after it’s clicked.**`
Use Axios from the CDN for the API call. (https://cdn.jsdelivr.net/npm/axios/dist/axios.min.js)
If the joke API returns an error display a message to try again in a few moments ❌✔ #implemented delay, but error isn't popping up
The punch line should not show immediately, but appear after 4 seconds ❌✔ #implemented delay, but delay either isn't happening, is too small, or our call isn't working

**`You can use the onClick() method to get ahold of your buttons. Seems like you’re on the right track with the timeout function so I won’t comment on the code. Is 400 supposed to be 4000 on line 75?`**

There should be a button to get another joke ✔
There should be <a> tag that links to the joke api documentation page ✔
This <a> should be manually styled as a button ✔ #done but need to fix border and sizing on ''
Buttons should have a background that is a gradient ✔❌  #currently on hover for both, joke-api is not gradient, but rest are 

`** ^^^ Add to menu buttons**`


All buttons (even the <a> that you made into a button) should have a hover effect  ✔

MarkCapka:
@tremorris1999 @schuywalker Thanks for the good feedback, fellas.

Will do some cleaning up - great suggestions all around!

TylerRose8:
Test Comment

Assignment7+8/css/site.css
```diff
@@ -0,0 +1,135 @@
...
+}
+
+.ddmenu{
+    position: fixed;
+    top: 5.5em;
+    right: 16%;
+    background-color: white;
+}
+
+.ddmenu button {
```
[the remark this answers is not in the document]
MarkCapka: fixed

Assignment7+8/index.html
MarkCapka: fixed

Assignment7+8/css/site.css
MarkCapka: fixed

Assignment7+8/css/site.css
MarkCapka: fixed

Assignment7+8/css/site.css
MarkCapka: fixed

Assignment7+8/css/site.css
MarkCapka: fixed

Assignment7+8/index.html
MarkCapka: style preference, also scripted in case we wanted to do anything with that button

Assignment7+8/js/index.js
MarkCapka: True, we do retrieve the joke from the API in getJoke then output it in telljoke, so they do have independent functionality. This also allows us to essentially queue a joke up for return as soon as the timer expires on the restriction of another api call.

Assignment7+8/js/index.js
MarkCapka: more granular access using getElementById paired with class, agree that onclick is probably an easier solution, but click works as well.

Assignment7+8/js/index.js
```diff
@@ -0,0 +1,108 @@
...
+        })
+        .catch(function (error) {
+            console.log("status code: " + error + " Try again in a few seconds..."); 
+        });
+}
+
+//Print the joke to the page
+function tellJoke() {
+    
+    let jokeSetup = jokeOutput.setup;
```
mmwoodfo, line 38: image
This line causes an error on load since your jokeOutput isn't set to anythig.

mmwoodfo (commented):
image
That's alot of files.

PR #160, opened 2022-03-19
Assignment9+10

sammyboy2000:
Initial PR.

### DIFF
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Assignment.Tests;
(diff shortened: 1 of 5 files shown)


Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -59,64 +60,85 @@ public void RunTaskAsync_Success()
...
+        result = task.Result;
+
         AssertValidPingOutput(result);
     }
 
     [TestMethod]
-#pragma warning disable CS1998 // Remove this
     async public Task RunAsync_UsingTpl_Success()
     {
         // DO use async/await in this test.
```
dhoxie, line 83: I would just remove this comment as the directions say to use async/await (like you do)

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -59,64 +60,85 @@ public void RunTaskAsync_Success()
...
-
         // Test Sut.RunAsync("localhost");
+        Task<PingResult> task = Sut.RunAsync("localhost");
+        result = await task;
         AssertValidPingOutput(result);
     }
-#pragma warning restore CS1998 // Remove this
-
 
+    //Having trouble getting this test to run on my setup.
```
dhoxie, line 91: This test runs successfully on my setup

dhoxie (changes requested):
## Review

4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for and array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ❌
   NOTE:
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay. ✔

## Extra Credit

## Fundamentals

- Ensure you enable:
  - net6 targeted ✔
  - C# 10.0 ✔
##
Looks great, I think you just need some work on async public Task RunAsync_MultipleHostAddresses_True().

PR #161, opened 2022-03-19
Assignment9+10

NicholasJohnsonDev:
Teammate Luis Garcia @TechGlue


sammyboy2000:
I will review this.

sammyboy2000:
## Instructions

   NOTE:
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. ❌

## Extra Credit

## Fundamentals

- Ensure you enable:
  - nullable reference types is enabled ❌
  - net6 targeted ❌
  - C# 10.0 ❌
- All of the above should be unit tested ❌

Looking great so far. All you need to do is enable your fundamentals in the csproj files and finish part 5.

NicholasJohnsonDev:
Those "fundementals" are done in the  https://github.com/IntelliTect-Samples/EWU-CSCD371-2022-Winter/blob/64b268c305432d4dcf8d5dd9a9cc99ce5509bd65/Assignment/Directory.Build.props file.

Question 5 was finished and some minor adjustments were made.
Made adjustments per warnings or suppressed them.

Assignment/Assignment/PingProcess.cs
```diff
@@ -13,54 +13,87 @@ public record struct PingResult(int ExitCode, string? StdOutput);
...
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        StartInfo.Arguments = hostNameOrAddress;
+        StringBuilder? stringBuilder = null;
+        void UpdateStdOutput(string? line) =>
+            (stringBuilder ??= new StringBuilder()).AppendLine(line);
+
+        Task<PingResult> task = Task.Run(
+            () =>
```
mmwoodfo, line 48: I would keep this on the line above

Assignment/Assignment/PingProcess.cs
```diff
@@ -13,54 +13,87 @@ public record struct PingResult(int ExitCode, string? StdOutput);
...
+            {
+                Process process = RunProcessInternal(StartInfo, UpdateStdOutput, default, cancellationToken);
+                cancellationToken.ThrowIfCancellationRequested();
```
mmwoodfo, line 51: You may want to do this on the line above

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -58,65 +59,90 @@ public void Run_CaptureStdOutput_Success()
...
     [TestMethod]
-#pragma warning disable CS1998 // Remove this
     async public Task RunAsync_UsingTpl_Success()
     {
         // DO use async/await in this test.
-        PingResult result = default;
-
-        // Test Sut.RunAsync("localhost");
+        Task<PingResult> task = Sut.RunAsync("localhost");
+        PingResult result = await task;
```
mmwoodfo, line 88: You would typically do these on the same line:
```suggestion
        PingResult task = await Sut.RunAsync("localhost");
```

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -58,65 +59,90 @@ public void Run_CaptureStdOutput_Success()
...
     [ExpectedException(typeof(TaskCanceledException))]
-    public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException()
+    public async Task RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException()
     {
-        // Use exception.Flatten()
+        //Create a new token
+        CancellationTokenSource cancellationTokenSource = new();
+
+        //Cancel Token
+        cancellationTokenSource.Cancel();
```
mmwoodfo, line 119: You should probably wait to cancel until after you've started RunAsync

Assignment/Assignment/PingProcess.cs
```diff
@@ -13,54 +13,87 @@ public record struct PingResult(int ExitCode, string? StdOutput);
...
+            => compiledString.Trim() + currentString.Result.StdOutput).Trim());
+
+        // int? lineCount = stringBuilder?.ToString().Split(Environment.NewLine).Length; // format stringBuilder to 40 lines
         return new PingResult(total, stringBuilder?.ToString());
     }
 
-    async public Task<PingResult> RunLongRunningAsync(
-        string hostNameOrAddress, CancellationToken cancellationToken = default)
+    //not passing tests idk if it's me or my OS
+    async public Task<PingResult> RunLongRunningAsync(string hostNameOrAddress, CancellationToken cancellationToken = default)
```
mmwoodfo, line 83: You shouldn't need to use async/await here

Assignment/Assignment/PingProcess.cs
```diff
@@ -13,54 +13,87 @@ public record struct PingResult(int ExitCode, string? StdOutput);
...
+        void UpdateStdOutput(string? line) =>
+            (stringBuilder ??= new StringBuilder()).AppendLine(line);
+        
+        Task<PingResult> task = Task.Factory.StartNew(() =>
+        {
+            Process process = RunProcessInternal(StartInfo, UpdateStdOutput, default, cancellationToken); 
+            return new PingResult(process.ExitCode, stringBuilder?.ToString());
+        }, cancellationToken, TaskCreationOptions.LongRunning, TaskScheduler.Current);
+        
+        return await task;
```
mmwoodfo, line 96: Just return your task
```suggestion
        return task;

```

mmwoodfo (commented):
Looks good!

PR #162, opened 2022-03-21
Assignment9+10 @hgill17

dcox2005:
Assignment9+10

### DIFF
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -90,7 +90,6 @@ public class PingProcessTests
         AssertValidPingOutput(result);
     }
 
-
     [TestMethod]
     [ExpectedException(typeof(AggregateException))]
     public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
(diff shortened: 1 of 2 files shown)


jmplett:
I will review

kelseymcmahon:
## Instructions

   NOTE:
   NOTE: This method does **NOT** use `Task.Run`. ✔

## Extra Credit

## Fundamentals

- Ensure you enable:
  - nullable reference types is enabled ❌
  - net6 targeted ❌
  - C# 10.0 ❌

Overall everything looked great! I liked how your group used the ExceptionDispatchInfo to throw the exception for the test to catch, we could not figure that out and did it with Assert. 

I would comment that in the RunAsync_UsingTaskReturn_Success() test method a cancellation token is created, sent, and then cancelled. I don't think it is necessary to cancel the task here and this could be dropped. 

Also, don't forget to edit the project files to include the fundamentals that are listed above.

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -3,6 +3,8 @@
 using System.Collections.Generic;
```
mmwoodfo, line 3: You should have nullability and analyzers turned on

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,40 +27,46 @@ public PingResult Run(string hostNameOrAddress)
...
 
     async public Task<PingResult> RunLongRunningAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
+        Task<PingResult> task = Task.Factory.StartNew(
+            () => Run(hostNameOrAddress), cancellationToken, TaskCreationOptions.LongRunning, TaskScheduler.Current);
         await task;
-        throw new NotImplementedException();
+        return task.Result;
```
mmwoodfo, line 69: You can just return the task here and remove the `async` from your method

PR #165, opened 2022-03-21
Assignment9+10 @ Disciple4GIT

nwnel-017:


TylerRose8:
Good job with this project. The biggest correction I see that you need to make is filling in your .csproj files with the analyzers and other properties. When you are throwing your inner exception from the aggregate exception, I would recommend not using the null forgiveness operator since if a different exception was thrown you may run into casting issues and or throwing null if the inner exception didn't exist. Instead, check that you have an inner exception of the type you expect, then throw it.


**Instructions**
1. Implement PingProcess' public Task<PingResult> RunTaskAsync(string hostNameOrAddress) ✔
 - First implement public void RunTaskAsync_Success() test method to test PingProcess.RunTaskAsync() using "localhost". ✔

2. Implement PingProcess' async public Task<PingResult> RunAsync(string hostNameOrAddress) ✔
 - First implement the public void RunAsync_UsingTaskReturn_Success() test method to test PingProcess.RunAsync() using "localhost" without using async/await. ✔
 - Also implement the async public Task RunAsync_UsingTpl_Success() test method to test PingProcess.RunAsync() using "localhost" but this time DO using async/await. ✔
 
3. Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔ Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.

4. Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:

5. Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run.✔


**Extra Credit**

**Fundamentals**
Ensure you enable:
nullable reference types is enabled ❌
net6 targeted ❌
C# 10.0 ❌
and enabled .NET analyzers for both projects ❌

Assignment/Assignment/PingProcess.cs
```diff
@@ -24,43 +24,56 @@ public PingResult Run(string hostNameOrAddress)
...
-
+    //2 & 3
     async public Task<PingResult> RunAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
-    }
+        cancellationToken.ThrowIfCancellationRequested();
```
mmwoodfo, line 39: It's not a huge deal - but I would recommend putting this within the Task itself as one of the first things that are done when the task starts up. It is cleaner

PR #166, opened 2022-03-22
Assignment9+10 + @NazariyPogo

Ashrom0:
Partner = @NazariyPogo -Nazariy Pogoriliy

### DIFF
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Assignment.Tests;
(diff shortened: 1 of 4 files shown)


MikePHaynes:
Will review.

MikePHaynes (commented):
Looks pretty good, you guys seem to have covered all of the requirements well enough. The only issue I encountered was that there were build errors for "NU1108 Cycle detected" for the csproj files so I had to make a throwaway project so that I could test your code, after that it ran fine. Other than that good work!

## Instructions

   NOTE:
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay.

## Extra Credit

## Fundamentals

- Ensure you enable:
  - net6 targeted ✔
  - C# 10.0 ✔

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -58,42 +60,54 @@ public void Run_CaptureStdOutput_Success()
...
-#pragma warning restore CS1998 // Remove this
-
 
     [TestMethod]
     [ExpectedException(typeof(AggregateException))]
     public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
     {
-        
+        CancellationTokenSource cancellationTokenSource = new();
+        cancellationTokenSource.Cancel();
```
MikePHaynes, line 89: The compiler recognized some unnecessary assignments here but your test doesn't pass without them. Maybe something you could try and fix.

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -58,42 +60,54 @@ public void Run_CaptureStdOutput_Success()
...
+        Task<PingResult> task = Sut.RunAsync("localhost", cancellationTokenSource.Token);
+        PingResult result = task.Result;
     }
```
MikePHaynes, line 92: The compiler recognized some unnecessary assignments here on this line and the one above but your test doesn't pass without them. Maybe something you could try and fix.

Chantony0131:
I will review this

Chantony0131:
Overall, Everything looks really good! Mike covered what I wanted to go over but besides that correction, everything should work out. 

[](https://github.com/IntelliTect-Samples/EWU-CSCD371-2022-Winter/tree/Assignment9+10#extra-credit)Instructions
Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:
Extra Credit

Assignment/Assignment.Tests/PingProcessTests.cs
mmwoodfo, line 89: I would recommend putting this below line 90. Cancel the task after it's started up, rather than sending the cancelation along with startup.

PR #167, opened 2022-03-22
Assignment9+10

kelseymcmahon:
Partners: @whitneybolar and @schuywalker


MarkCapka:
doing peer review on this pull request, will comment with review shortly.

MarkCapka:
# EWU-CSCD371-2021-Winter

## See [Docs](Docs)

## Assignment 9&10

Peer Reviewer:  Mark Capka 
Overall notes:
Don't forget to add analyzers to your Assignment and Assigment.Tests '.csproj' files: 
	<PackageReference Include="IntelliTect.Analyzers" Version="0.1.8" />
**can then clean up/suppress the errors in the test files for PingProcessTests since it'll be mad about underscores and PascalCase, but may catch some other convention stuff. 

#3 - see the PR note, I think that is more what they were looking for although I do think that the way that you wrote it does technically fulfill the requirements as described, it doesn't work with the decorator. Should be a quick and easy fix though.

Other than that, excellent work! I think you hit all the check marks. I did make a couple stylistic and simplcity comments, but I don't know that it's anything aggregious enough for points off. 

Great quarter with all ya'll in class!




1. Implement `PingProcess`' `public Task<PingResult> RunTaskAsync(string hostNameOrAddress)` ✔ PR NOTE: -- confirmed
   - First implement `public void RunTaskAsync_Success()` test method to test `PingProcess.RunTaskAsync()` using `"localhost"`. ✔PR NOTE: -- confirmed
   - Do NOT use async/await in this implementation. ✔ PR NOTE: -- confirmed uses Task.Wait(), although i'm not positive that you need it. (take that note with a grain of salt though)
2. Implement `PingProcess`' `async public Task<PingResult> RunAsync(string hostNameOrAddress)` ✔PR NOTE: -- confirmed
   - First implement the `public void RunAsync_UsingTaskReturn_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` **without** using async/await. ✔ PR NOTE: -- confirmed
   - Also implement the `async public Task RunAsync_UsingTpl_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` but this time **DO** using async/await. ✔PR NOTE: -- confirmed
3. Add support for an optional cancellation token to the `PingProcess.RunAsync()` signature. ✔ ✔PR NOTE: -- confirmed
   Inside the `PingProcess.RunAsync()` invoke the token's `ThrowIfCancellationRequested()` method so an exception is thrown. ✔✔PR NOTE: -- confirmed
   Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ✔✔PR NOTE: -- confirmed
 with a `TaskCanceledException` inner exception X        PR NOTE: -- seems to only be throwing an AggregateException if the method is decorated with the TaskCanceledException. replace the Assert statement in the catch with an if(e.InnerException is not null) throw e.InnerException. Decorate it back up and it'll throw. 
 using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping`  ✔PR NOTE: -- confirmed
and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ ✔PR NOTE: -- confirmed
respectively.
4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for and array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ✔✔PR NOTE: -- confirmed
   NOTE:
      - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔ ✔PR NOTE: -- confirmed
   NOTE: This method does **NOT** use `Task.Run`. ✔PR NOTE: -- confirmed    
              ----NOTE: another **heavy** grain of salt here, but i'm not sure that you need all of the setup with the getPingResult in here. Test passes though and seems to be working correclty. 

## Extra Credit

## Fundamentals

- Ensure you enable:
  - net6 targeted ✔
  - C# 10.0 ✔
  - and enabled .NET analyzers for both projects ❌  --PR NOTE: **both** projects, need to add to Assignment.csproj
- Choose simplicity over complexity ✔  PR NOTE: overall, I think yes, May have some extraneous stuff in the RunLongRunningAsync, but i'm honestly not sure that your implementation isn't better than the approach we had.

tremorris1999:
I will also review this.

tremorris1999 (approved):
## Instructions

   Test that, when cancelled from the test method, the exception thrown is an `AggregateException`✔ with a `TaskCanceledException` inner exception✔ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ✔and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ respectively.
   NOTE:
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay.✔

## Extra Credit

## Fundamentals

- Ensure you enable:
  - net6 targeted ✔
  - C# 10.0✔

---
Overall, I think it looks damn good. I don't have any recommendations for changes here. You guys did a heck of a job here!

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,40 +27,63 @@ public PingResult Run(string hostNameOrAddress)
 
     public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
     {
-        throw new NotImplementedException();
+        Task<PingResult> task = Task.Run(() => Run(hostNameOrAddress)
+            );
+
+        task.Wait();
+        return task;
```
mmwoodfo, line 34: You can just return the task here
```suggestion
        return task;
```

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,40 +27,63 @@ public PingResult Run(string hostNameOrAddress)
...
     }
 
-    async public Task<PingResult> RunAsync(
-        string hostNameOrAddress, CancellationToken cancellationToken = default)
+    async public Task<PingResult> RunAsync(string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        cancellationToken.ThrowIfCancellationRequested();
```
mmwoodfo, line 39: I would recommend configuring the token within the task body.

Something like this:
```
Task<PingResult> task = Task.Run(() =>
        {
            cancellationToken.ThrowIfCancellationRequested();
            ...
        });
```

PR #168, opened 2022-03-22
Assignment9+10 @ZackWard-student

jmplett:
@ZackWard-student


TechGlue:
I'll review this!

TechGlue (commented):
1. Implement `PingProcess`' `public Task<PingResult> RunTaskAsync(string hostNameOrAddress)` ✔
    - First implement `public void RunTaskAsync_Success()` test method to test `PingProcess.RunTaskAsync()` using `"localhost"`. ✔
    - Do NOT use async/await in this implementation. ✔
2. Implement `PingProcess`' `async public Task<PingResult> RunAsync(string hostNameOrAddress)` ✔
    - First implement the `public void RunAsync_UsingTaskReturn_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` **without** using async/await. ✔
    - Also implement the `async public Task RunAsync_UsingTpl_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` but this time **DO** using async/await. ✔
3. Add support for an optional cancellation token to the `PingProcess.RunAsync()` signature. ✔ Inside the `PingProcess.RunAsync()` invoke the token's `ThrowIfCancellationRequested()` method so an exception is thrown. ✔ Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ✔with a `TaskCanceledException` inner exception ✔ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ✔and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ respectively.
4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for and array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ✔ NOTE:
    - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay. NOTE: This method does **NOT** use `Task.Run`.✔

TechGlue:
Other than your .csproj files missing some of the fundamentals everything else looks good. Great Work!

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,40 +27,61 @@ public PingResult Run(string hostNameOrAddress)
...
-        throw new NotImplementedException();
+        Task<PingResult> task = Task.Run(() => Run(hostNameOrAddress));
+        return task;
     }
 
     async public Task<PingResult> RunAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
+        cancellationToken.ThrowIfCancellationRequested();
```
mmwoodfo, line 37: I would recommend putting this within the task that is running - but it isn't a big deal.

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -67,56 +72,75 @@ public void RunAsync_UsingTaskReturn_Success()
...
 
 
     [TestMethod]
     [ExpectedException(typeof(AggregateException))]
     public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
     {
-        
+        CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+        CancellationToken token = cancellationTokenSource.Token;
+        cancellationTokenSource.Cancel();
```
mmwoodfo, line 99: I would recommend calling this after the task starts, rather than beforehand so you can make sure the task cancels anytime during its run rather than during its start-up.

PR #170, opened 2022-03-22
Assignment9+10

MarkCapka:
@TylerRose8

 Thanks for a great quarter, ya'll.


ZackWard-student:
Peer reviewing.

ZackWard-student:
1. Implement PingProcess' public Task<PingResult> RunTaskAsync(string hostNameOrAddress) ✔
- First implement public void RunTaskAsync_Success() test method to test PingProcess.RunTaskAsync() using "localhost". ✔

2. Implement PingProcess' async public Task<PingResult> RunAsync(string hostNameOrAddress) ✔

- First implement the public void RunAsync_UsingTaskReturn_Success() test method to test PingProcess.RunAsync() using "localhost" without using async/await. ✔
- Also implement the async public Task RunAsync_UsingTpl_Success() test method to test PingProcess.RunAsync() using "localhost" but this time DO using async/await. ✔

3. Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔

- Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔ 
- Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔
- with a TaskCanceledException inner exception ✔
- using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔
- and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively

4. Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ 

5. Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run.✔

All the Basic stuff looks goo too!

Assignment/Assignment/PingProcess.cs
```diff
@@ -9,63 +9,188 @@
...
+        cancellationToken.ThrowIfCancellationRequested();
+        return taskPing.Result;
+
     }
 
-    async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
-    {
-        StringBuilder? stringBuilder = null;
-        ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
+    async public Task<PingResult> RunAsync(System.IProgress<PingResult> progress, CancellationToken cancellationToken = default, params string[] hostNameOrAddresses)
```
mmwoodfo, line 136: You should have more tests around the extra credit implementation to ensure that it is working as expected.

mmwoodfo:
You've got alot of additional files in your git diff that don't need to be there.

PR #171, opened 2022-03-22
Assignment9+10

xiokd:
Code Pair: @dhoxie

### DIFF
--- a/Assignment/Assignment.Tests/Assignment.Tests.csproj
+++ b/Assignment/Assignment.Tests/Assignment.Tests.csproj
@@ -2,6 +2,10 @@
 
   <PropertyGroup>
     <IsPackable>false</IsPackable>
+    <TargetFramework>net6.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <LangVersion>10.0</LangVersion>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -73,21 +73,18 @@ public class PingProcessTests
     }
 
     [TestMethod]
-#pragma warning disable CS1998 // Remove this
     async public Task RunAsync_UsingTpl_Success()
     {
-        // DO use async/await in this test.
-        PingResult result = await Sut.RunAsync("localhost");
+        PingResult result = default;
+        Task<PingResult> task = Sut.RunAsync("localhost");
+        result = await task;
 
-        // Test Sut.RunAsync("localhost");
         AssertValidPingOutput(result);
     }
-#pragma warning restore CS1998 // Remove this
-
 
     [TestMethod]
     [ExpectedException(typeof(AggregateException))]
-    public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
+    public async void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
     {
         CancellationTokenSource cancellationTokenSource = new();
         CancellationToken cancellationToken = cancellationTokenSource.Token;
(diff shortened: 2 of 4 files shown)


DakotaCondos:
I will leave a review for this

DakotaCondos:
Instructions
Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔ 

> (I think you did this section very well, it looks nicer than our implementation) 

Test that, when canceled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.
Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ❌&✔ 

> (you're adding the entirety of the process output to the stringbuilder, not each individual tasks.StdOutput. Here is how we implemented it, don't forget the .trim() !!!)

```
async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
    {
        StringBuilder stringBuilder = new();
        ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
        {
            Task<PingResult> task = Task.Run(() => Run(item));
            await task.WaitAsync(default(CancellationToken));
            stringBuilder.AppendLine(task.Result.StdOutput);
            return task.Result.ExitCode;
        });
 
        await Task.WhenAll(all);
        int total = all.Aggregate(0, (total, item) => total + 1);
        return new PingResult(total, stringBuilder?.ToString().Trim());
    }
```
 
NOTE:
The order of the items in the stdOutput is irrelevant and expected to be intermingled.
StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌&✔
Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run. ❌(not yet implemented)
Extra Credit
Test and implement PingProcess.RunAsync(System.IProgress<T> progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌(not yet implemented)

schuywalker:
I will review 2

schuywalker:
Instructions
Also implement the async public Task RunAsync_UsingTpl_Success() test method to test PingProcess.RunAsync() using "localhost" but this time DO using async/await. ❌ `- implemented but not using async await`
Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔ Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping  `maybe try using a Try Catch block instead of expected exception? The test wont run for me.` ✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ❌ respectively. 
Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ❌ NOTE:
StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌
Extra Credit
Fundamentals
Ensure you enable:
net6 targeted ✔
C# 10.0 ✔

PR #172, opened 2022-03-22
Assignment9+10

mpennock5:


czamora7 (commented):

Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔ Test that, when cancelled from the test method, the exception thrown is an AggregateException ❌ 
with a TaskCanceledException inner exception ❌ the method will throw an OperationCanceledException
using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.

Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:
StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌ This needs to be tested for

❌ the method header does not take the specified parameters and RunProcessInternal is not invoked

Extra Credit
Test and implement PingProcess.RunAsync(System.IProgress<T> progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌ not attempted

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -59,42 +61,68 @@ public void RunTaskAsync_Success()
...
+        Task<PingResult> task = Sut.RunAsync("localhost", token);
+
+        try
+        {
+            PingResult result = task.Result;
+        }
+        catch (AggregateException exception)
+        {
+            ExceptionDispatchInfo.Capture(exception.Flatten().InnerException!).Throw();
+        }
```
tnadalini, line 125: not sure if the try catch is necessary, but otherwise this looks like a good test

tnadalini (commented):
    Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:

Extra Credit

Fundamentals

    Ensure you enable:
        net6 targeted ✔
        C# 10.0 ✔

mmwoodfo:
Did you do a PR review? @mpennock5

PR #173, opened 2022-03-22
Assignment9+10 @ericlplanteschool

MikePHaynes:
Partner: @ericlplanteschool


ericlplanteschool:
reviewed @GavenElder

A-Tanner:
I will review this

NazariyPogo:
Will review

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -38,63 +41,89 @@ public void Run_GoogleDotCom_Success()
...
+
+            cancellationTokenSource.Cancel();
+
+            Task<PingResult> task = Sut.RunAsync("localhost", cancellationToken);
+
+            PingResult result = task.Result;
+        }
+        catch (AggregateException aggregateException)
+        {
+            foreach (var exception in aggregateException.Flatten().InnerExceptions)
```
A-Tanner, line 115: You have to throw this exception in the RunAsync method. This test does not pass because the method throws no aggregate exception.

A-Tanner (commented):
Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ❌ --> Use the above described to throw an exception in RunAsync
Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ❌✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.
Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ❌--> Based on the test time, I don't think this is properly asynchronous.

 NOTE:
Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run. ❌✔ --> Test does not use TPL (the name suggests it should)

Ensure you enable:
net6 targeted ✔
C# 10.0 ✔




Overall, pretty good but needs a bit more TLC. Major complaints are that the IEnumerable method didn't appear to be properly asynchronous, RunAsync does not throw the required exception if token is canceled, and RunLongRunningAsync does not use the TPL library for testing.

NazariyPogo:
   Inside the `PingProcess.RunAsync()` invoke the token's `ThrowIfCancellationRequested()` method so an exception is thrown. ❌ I don't see this thrown anywhere in your methods.
   Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ✔ with a `TaskCanceledException` inner exception ✔ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ❌Not implemented
and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ respectively.
   NOTE:
   NOTE: This method does **NOT** use `Task.Run`.❌✔Check Tanner's comment above

## Extra Credit

## Fundamentals

- Ensure you enable:
  - net6 targeted ✔
  - C# 10.0 ✔

Looks pretty great overall, just missing the cancellation token implementation (Unless I'm blind) and the one test method that you couldn't figure out.

MikePHaynes:
@NazariyPogo @A-Tanner Helpful feedback, thanks guys!

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,40 +29,49 @@ public PingResult Run(string hostNameOrAddress)
...
-        ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
-        {
-            Task<PingResult> task = null!;
-            // ...
-
-            await task.WaitAsync(default(CancellationToken));
-            return task.Result.ExitCode;
-        });
+     
+        ParallelQuery<Task<PingResult>>? all = hostNameOrAddresses.AsParallel()
```
mmwoodfo, line 53: You should make sure that you are setting the arguments to the item you are selecting here. 
`StartInfo.Arguments = item;`

You should also be calling `RunProcessInternal` somewhere in here.

PR #175, opened 2022-03-22
Assignment 9+10

GavenElder:


ericlplanteschool:
will review

ericlplanteschool:
1.  Implement PingProcess' public Task<PingResult> RunTaskAsync(string hostNameOrAddress) ✔

2.  Implement PingProcess' async public Task<PingResult> RunAsync(string hostNameOrAddress) ✔

3.  Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔ 
Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.

4. Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:

5.  Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run.

ericlplanteschool (commented):
implement #5 otherwise lgtm

htran9:
I will review this.

htran9 (commented):
Overall it looks pretty good.
There are a few things that I would suggest to make changes to.
1. For your RunAsync multiplehostaddresses, you would need to concenate the output before returning it, so that way you will get the proper output to test.
2. There are a few print statements that I saw in the test file that should probably be removed prior to the final submission. 

Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ 
NOTE:
StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌(Not returning all the process output)

Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run. ❌
(Did not see an implementation for this method.)

Extra Credit

Test and implement PingProcess.RunAsync(System.IProgress<T> progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌(optional, did not see an implementation.)

Fundamentals

Ensure you enable:
nullable reference types is enabled ❌(Did not see this added.)
net6 targeted ❌(Did not see this added.)
C# 10.0 ❌(Did not see this added.)
and enabled .NET analyzers for both projects ❌(Did not see this added.)
All of the above should be unit tested ❌(Did not see this added.)

nwnel-017:
Since every pull request has been reviewed, I will review this one

nwnel-017:
Overall it looks pretty good. You still need to implement step 5 but the rest of it looks good for the most part. you definitely need to add the extra specifications in the project file to enable nullability, net 6, and enable analyzers for the project. You also need to finish your testing, it looks like not all of the tests are finished

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,40 +27,38 @@ public PingResult Run(string hostNameOrAddress)
...
 
-    async public Task<PingResult> RunLongRunningAsync(
-        string hostNameOrAddress, CancellationToken cancellationToken = default)
+     public Task<PingResult> RunLongRunningAsync(ProcessStartInfo startInfo,
+        Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        //still work in progress
```
mmwoodfo, line 61: It would be better to use `throw new NotImplementedException()` so the code is runnable.

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -57,66 +59,80 @@ public void Run_CaptureStdOutput_Success()
...
-
 
     [TestMethod]
     [ExpectedException(typeof(AggregateException))]
     public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
     {
-        
+        CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+        CancellationToken token = cancellationTokenSource.Token;
+        cancellationTokenSource.Cancel();
```
mmwoodfo, line 87: You should cancel the token after you start the task.

mmwoodfo (commented):
Your code doesn't compile. When fixing the compilation issues some of your tests are failing.

